PR #15, opened 2025-04-30
Update sample code to match imports and function usage

jtemporal:
This updates the README on `auth0-fastapi` so the code sample works in a copy-and-paste scenario:

1. adds missing `register_auth_routes` in the minimal example
2. fixes the `register_auth_routes` usage in both examples on the readme passing the config and router variables individually instead of a tuple


cristiandouce (dismissed):

kishore7snehil (approved):
LGTM!

PR #17, opened 2025-05-01
Adds example with route protection via path decorator in API

jtemporal:
This adds an example with the usage of the `require_auth()` method in path operation decorator.


kishore7snehil (approved):
LGTM!

PR #22, opened 2025-07-28
feat: Add DPoP authentication support

kishore7snehil:
### 📋 Changes

This PR implements DPoP (RFC 9449) support for auth0-api-python, enabling sender-constrained OAuth 2.0 access tokens using cryptographic proof of possession for enhanced API security.

> [!NOTE]
> DPoP is currently in Early Access. Contact Auth0 support to enable it for your tenant.

#### ✨ Features
- **DPoP Protocol Implementation:** Complete RFC 9449 compliant implementation with ES256 signature verification, JWT proof validation, and comprehensive claim checking
- **Dual Authentication Scheme Support:** Seamless handling of both Bearer and DPoP authentication schemes in a single API
- **Unified Entry Point:** New `verify_request()` method automatically detects and validates Bearer or DPoP schemes
- **Flexible Configuration Modes:** Support for "Allowed Mode" (mixed Bearer/DPoP) and "Required Mode" (DPoP-only enforcement)
- **Comprehensive Validation:** Full claim validation including `htm`, `htu`, `ath`, `jti` with timing checks and replay protection
- **URL Normalization:** RFC-compliant URL matching for DPoP `htu` claims with proper scheme and port handling

#### 🔧 API Changes
- Added `verify_request()` method to `ApiClient` for unified authentication scheme detection
- Added `verify_dpop_proof()` method for direct DPoP proof verification
- Extended `ApiClientOptions` with DPoP configuration parameters (`dpop_enabled`, `dpop_required`, timing settings)
- New DPoP-specific error classes: `InvalidDpopProofError`, `InvalidAuthSchemeError`
- Enhanced error handling with proper HTTP status codes and WWW-Authenticate headers

#### 📚 Core Components
- **JWT Dual Support:** Separate JsonWebToken handlers for RS256 (Bearer) and ES256 (DPoP) validation
- **Cryptographic Utilities:** SHA-256 hashing, Base64URL encoding, JWK thumbprint calculation
- **URL Processing:** DPoP-compliant URL normalization and comparison
- **Error Response System:** HTTP-aware error hierarchy with OAuth error codes and WWW-Authenticate headers
- **Test Infrastructure:** Comprehensive utilities for generating DPoP proofs and bound tokens

#### 📖 Documentation
- Updated `README.md` with comprehensive features overview and DPoP authentication section
- Added detailed configuration examples for both "Allowed" and "Required" modes
- Created `DPOP.md` with complete integration examples

### 📎 References

- [RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html)

#### 🧪 Testing
- **Comprehensive Test Suite:** 35+ tests covering all DPoP validation scenarios organized by claim types (IAT, HTU, JTI)
- **Test Utilities:** Built-in `generate_dpop_proof()` and `generate_token_with_cnf()` for development and testing
- **Error Scenario Coverage:** Complete testing of validation failures, timing edge cases, and malformed proofs
- **Integration Testing:** Full request verification flow testing for both authentication schemes

### Contributor Checklist

### DIFF
--- a/packages/auth0_api_python/README.md
+++ b/packages/auth0_api_python/README.md
@@ -100,6 +100,9 @@ If the token lacks `my_custom_claim` or fails any standard check (issuer mismatc
 
 ### 4. DPoP Authentication
 
+> [!NOTE]  
+> This feature is currently available in [Early Access](https://auth0.com/docs/troubleshoot/product-lifecycle/product-release-stages#early-access). Please reach out to Auth0 support to get it enabled for your tenant.
+
 This library supports **DPoP (Demonstrating Proof-of-Possession)** for enhanced security, allowing clients to prove possession of private keys bound to access tokens.
 
 #### Allowed Mode (Default)
--- a/packages/auth0_api_python/README.md
+++ b/packages/auth0_api_python/README.md
@@ -6,6 +6,24 @@ It’s intended as a foundation for building more framework-specific integration
 
 📚 [Documentation](#documentation) - 🚀 [Getting Started](#getting-started) - 💬 [Feedback](#feedback)
 
+## Features & Authentication Schemes
+
+This SDK provides comprehensive support for securing APIs with Auth0-issued access tokens:
+
+### **Authentication Schemes**
+- **Bearer Token Authentication** - Traditional OAuth 2.0 Bearer tokens (RS256)
+- **DPoP Authentication** - Enhanced security with Demonstrating Proof-of-Possession (ES256)
+- **Mixed Mode Support** - Seamlessly handle both Bearer and DPoP in the same API
+
+### **Core Features**
+- **Unified Entry Point**: `verify_request()` - automatically detects and validates Bearer or DPoP schemes
+- **OIDC Discovery** - Automatic fetching of Auth0 metadata and JWKS
+- **JWT Validation** - Complete RS256 signature verification with claim validation
+- **DPoP Proof Verification** - Full RFC 9449 compliance with ES256 signature validation
+- **Flexible Configuration** - Support for both "Allowed" and "Required" DPoP modes
+- **Comprehensive Error Handling** - Detailed errors with proper HTTP status codes and WWW-Authenticate headers
+- **Framework Agnostic** - Works with FastAPI, Django, Flask, or any Python web framework
+
 ## Documentation
 
 - [Docs Site](https://auth0.com/docs) - explore our docs site and learn more about Auth0.
@@ -79,6 +97,60 @@ decoded_and_verified_token = await api_client.verify_access_token(
 ```
 
 If the token lacks `my_custom_claim` or fails any standard check (issuer mismatch, expired token, invalid signature), the method raises a `VerifyAccessTokenError`.
+
+### 4. DPoP Authentication
+
+This library supports **DPoP (Demonstrating Proof-of-Possession)** for enhanced security, allowing clients to prove possession of private keys bound to access tokens.
+
+#### Allowed Mode (Default)
+
+Accepts both Bearer and DPoP tokens - ideal for gradual migration:
+
+```python
+api_client = ApiClient(ApiClientOptions(
+    domain="<AUTH0_DOMAIN>",
+    audience="<AUTH0_AUDIENCE>",
+    dpop_enabled=True,      # Default - enables DPoP support
+    dpop_required=False     # Default - allows both Bearer and DPoP
+))
+
+# Use verify_request() for automatic scheme detection
+result = await api_client.verify_request(
+    headers={
+        "authorization": "DPoP eyJ0eXAiOiJKV1Q...",  # DPoP scheme
+        "dpop": "eyJ0eXAiOiJkcG9wK2p3dC...",        # DPoP proof
+    },
+    http_method="GET",
+    http_url="https://api.example.com/resource"
+)
+```
+
+#### Required Mode
+
+Enforces DPoP-only authentication, rejecting Bearer tokens:
+
+```python
+api_client = ApiClient(ApiClientOptions(
+    domain="<AUTH0_DOMAIN>",
+    audience="<AUTH0_AUDIENCE>", 
+    dpop_required=True      # Rejects Bearer tokens
+))
+```
+
+#### Configuration Options
+
+```python
+api_client = ApiClient(ApiClientOptions(
+    domain="<AUTH0_DOMAIN>",
+    audience="<AUTH0_AUDIENCE>",
+    dpop_enabled=True,          # Enable/disable DPoP support
+    dpop_required=False,        # Require DPoP (reject Bearer)
+    dpop_iat_leeway=30,         # Clock skew tolerance (seconds)
+    dpop_iat_offset=300,        # Maximum proof age (seconds)
+))
+```
+
+📖 **[Complete DPoP Documentation](docs/DPOP.md)** - Detailed guide with examples, error handling, and security considerations.
 
 ## Feedback
 
--- a/packages/auth0_api_python/src/auth0_api_python/api_client.py
+++ b/packages/auth0_api_python/src/auth0_api_python/api_client.py
@@ -1,11 +1,20 @@
 import time
-from typing import Optional, List, Dict, Any
+import hashlib
+from typing import Optional, List, Dict, Any, Tuple
 
 from authlib.jose import JsonWebToken, JsonWebKey
 
 from .config import ApiClientOptions
-from .errors import MissingRequiredArgumentError, VerifyAccessTokenError
-from .utils import fetch_oidc_metadata, fetch_jwks, get_unverified_header
+from .errors import (
+    MissingRequiredArgumentError, 
+    VerifyAccessTokenError, 
+    InvalidAuthSchemeError, 
+    InvalidDpopProofError,
+    MissingAuthorizationError,  
+    BaseAuthError
+)
+from .utils import fetch_oidc_metadata, fetch_jwks, get_unverified_header, normalize_url_for_htu, sha256_base64url, calculate_jwk_thumbprint
+
 
 class ApiClient:
     """
@@ -14,7 +23,6 @@ class ApiClient:
     """
 
     def __init__(self, options: ApiClientOptions):
-
         if not options.domain:
             raise MissingRequiredArgumentError("domain")
         if not options.audience:
@@ -26,25 +34,185 @@ class ApiClient:
 
         self._jwt = JsonWebToken(["RS256"])
 
-    async def _discover(self) -> Dict[str, Any]:
-        """Lazy-load OIDC discovery metadata."""
-        if self._metadata is None:
-            self._metadata = await fetch_oidc_metadata(
-                domain=self.options.domain,
-                custom_fetch=self.options.custom_fetch
-            )
-        return self._metadata
+        self._dpop_algorithms = ["ES256"]
+        self._dpop_jwt = JsonWebToken(self._dpop_algorithms)
 
-    async def _load_jwks(self) -> Dict[str, Any]:
-        """Fetches and caches JWKS data from the OIDC metadata."""
-        if self._jwks_data is None:
-            metadata = await self._discover()
-            jwks_uri = metadata["jwks_uri"]
-            self._jwks_data = await fetch_jwks(
-                jwks_uri=jwks_uri, 
-                custom_fetch=self.options.custom_fetch
+    
+    async def verify_request(
+        self,
+        headers: Dict[str, str],
+        http_method: Optional[str] = None,
+        http_url: Optional[str] = None
+    ) -> Dict[str, Any]:
+        """
+        Dispatch based on Authorization scheme:
+          • If scheme is 'DPoP', verifies both access token and DPoP proof
+          • If scheme is 'Bearer', verifies only the access token
+
+        Args:
+            headers: HTTP headers dict containing:
+                - "authorization": The Authorization header value (required)
+                - "dpop": The DPoP proof header value (required for DPoP)
+            http_method: The HTTP method (required for DPoP)
+            http_url: The HTTP URL (required for DPoP)
+
+        Returns:
+            The decoded access token claims
+
+        Raises:
+            MissingRequiredArgumentError: If required args are missing
+            InvalidAuthSchemeError: If an unsupported scheme is provided
+            InvalidDpopProofError: If DPoP verification fails
+            VerifyAccessTokenError: If access token verification fails
+        """
+        authorization_header = headers.get("authorization", "")
+        dpop_proof = headers.get("dpop")
+
+        if not authorization_header:
+            if getattr(self.options, "dpop_required", False):
+                raise self._prepare_error(
+                        InvalidAuthSchemeError("Expecting Authorization header with DPoP scheme.")
+                    )
+            else :
+                raise self._prepare_error(MissingAuthorizationError())
+        
+        
+        parts = authorization_header.split(" ", 1)
+        if len(parts) < 2:
+            raise self._prepare_error(MissingAuthorizationError())
+       
+        
+        try:
+            scheme, token = authorization_header.split(" ", 1)
+        except ValueError:
+            raise self._prepare_error(
+                MissingAuthorizationError()
             )
-        return self._jwks_data
+        
+
+        scheme = scheme.strip().lower()
+
+        if getattr(self.options, "dpop_required", False) and scheme != "dpop":
+                if scheme == "bearer":
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError("Invalid scheme. Expected 'DPoP', but got 'bearer'."),
+                        auth_scheme=scheme
+                    )
+                else:
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError("Invalid scheme. Expected 'DPoP' scheme."),
+                        auth_scheme=scheme
+                    )
+        if not token.strip():
+            raise self._prepare_error(MissingAuthorizationError())
+
+
+        if scheme == "dpop":
+            if not self.options.dpop_enabled:
+                raise self._prepare_error(MissingAuthorizationError())
+            
+            if not dpop_proof:
+                if getattr(self.options, "dpop_required", False):
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError("Expecting Authorization header with DPoP scheme."),
+                        auth_scheme=scheme
+                    )
+                else:
+                    raise self._prepare_error(
+                        InvalidDpopProofError("Operation indicated DPoP use but the request has no DPoP HTTP Header"), 
+                        auth_scheme=scheme
+                    )
+            
+            if "," in dpop_proof:
+                raise self._prepare_error(
+                    InvalidDpopProofError("Multiple DPoP proofs are not allowed"),
+                    auth_scheme=scheme
+                )
+
+            try:
+                await get_unverified_header(dpop_proof)
+            except Exception:
+                raise self._prepare_error(InvalidDpopProofError("Failed to verify DPoP proof"), auth_scheme=scheme)
+            
+            if not http_method or not http_url:
+                raise self._prepare_error(
+                    InvalidDpopProofError("Operation indicated DPoP use but the request has no http_method or http_url"), auth_scheme=scheme
+                )
+            
+            try:
+                access_token_claims = await self.verify_access_token(token)
+            except VerifyAccessTokenError as e:
+                raise self._prepare_error(e, auth_scheme=scheme)
+            
+            cnf_claim = access_token_claims.get("cnf")
+            
+            if not cnf_claim:
+                raise self._prepare_error(
+                    InvalidDpopProofError("Operation indicated DPoP use but the JWT Access Token has no jkt confirmation claim"),
+                    auth_scheme=scheme
+                )
+            
+            if not isinstance(cnf_claim, dict):
+                raise self._prepare_error(
+                    InvalidDpopProofError("Operation indicated DPoP use but the JWT Access Token has invalid confirmation claim format"),
+                    auth_scheme=scheme
+                )
+            try:
+                await self.verify_dpop_proof(
+                    access_token=token,
+                    proof=dpop_proof,
+                    http_method=http_method,
+                    http_url=http_url
+                )
+            except InvalidDpopProofError as e:
+                raise self._prepare_error(e, auth_scheme=scheme)
+
+            # DPoP binding verification
+            jwk_dict = (await get_unverified_header(dpop_proof))["jwk"]
+            actual_jkt = calculate_jwk_thumbprint(jwk_dict)
+            expected_jkt = cnf_claim.get("jkt")
+
+            if not expected_jkt:
+                raise self._prepare_error(
+                    VerifyAccessTokenError("Access token 'cnf' claim missing 'jkt'"),
+                    auth_scheme=scheme
+                )
+            
+            if expected_jkt != actual_jkt:
+                raise self._prepare_error(
+                    VerifyAccessTokenError("JWT Access Token confirmation mismatch"),
+                    auth_scheme=scheme
+                )
+            
+            return access_token_claims
+
+        if scheme == "bearer": 
+            if dpop_proof:
+                if self.options.dpop_enabled:
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError(
+                            "Operation indicated DPoP use but the request's Authorization HTTP Header scheme is not DPoP"
+                        ),
+                        auth_scheme=scheme
+                    )
+                
+            try:
+                claims = await self.verify_access_token(token)
+                if claims.get("cnf") and claims["cnf"].get("jkt"):
+                    if self.options.dpop_enabled:
+                        raise self._prepare_error(
+                            InvalidAuthSchemeError(
+                                "Operation indicated DPoP use but the request's Authorization HTTP Header scheme is not DPoP"
+                            ),
+                            auth_scheme=scheme
+                        )
+                        
+                
+                return claims
+            except VerifyAccessTokenError as e:
+                raise self._prepare_error(e, auth_scheme=scheme)
+
+        raise self._prepare_error(MissingAuthorizationError())
 
     async def verify_access_token(
         self,
@@ -71,7 +239,6 @@ class ApiClient:
 
         required_claims = required_claims or []
 
-    
         try:
             header = await get_unverified_header(access_token)
             kid = header["kid"]
@@ -100,7 +267,6 @@ class ApiClient:
         metadata = await self._discover()
         issuer = metadata["issuer"]
 
-
         if claims.get("iss") != issuer:
             raise VerifyAccessTokenError("Issuer mismatch")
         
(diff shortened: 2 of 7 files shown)


.github/workflows/test-auth0-api-python.yml
```diff
@@ -0,0 +1,63 @@
+name: Test auth0-api-python
+
+on:
+  push:
+    branches:
+      - feature/auth0-api-python
```
Widcket, line 6: This will only apply to a single branch called `feature/auth0-api-python`. Is this intended?
Though, is this rule necessary given there is one using paths below?

kishore7snehil: Yes, this is intended right now. This repo will be pointing out to only `auth0-server-python` SDK/Library in coming few weeks. We have the new `auth0-api-python` repo created. That will have the relevant code.

kishore7snehil: I have added this workflow, just to give visibility on the tests.

.github/workflows/test-auth0-api-python.yml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Install dependencies
+      if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+      working-directory: ./packages/auth0_api_python
+      run: poetry install --no-interaction --no-root
+    
+    - name: Install package
+      working-directory: ./packages/auth0_api_python
+      run: poetry install --no-interaction
+    
+    - name: Run tests with pytest
```
Widcket, line 55: I'd suggest also running the tests before building and releasing (https://github.com/auth0/auth0-server-python/blob/main/.github/workflows/release.yml#L51).

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+        http_method: Optional[str] = None,
+        http_url: Optional[str] = None
+    ) -> dict[str, Any]:
+        """
+        Dispatch based on Authorization scheme:
+          • If scheme is 'DPoP', verifies both access token and DPoP proof
+          • If scheme is 'Bearer', verifies only the access token
+
+        Args:
+            headers: HTTP headers dict containing:
```
Widcket, line 59: This should mention that the headers must have the keys lowercased.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+            The decoded access token claims
+
+        Raises:
+            MissingRequiredArgumentError: If required args are missing
+            InvalidAuthSchemeError: If an unsupported scheme is provided
+            InvalidDpopProofError: If DPoP verification fails
+            VerifyAccessTokenError: If access token verification fails
+        """
+        authorization_header = headers.get("authorization", "")
+        dpop_proof = headers.get("dpop")
```
Widcket, line 75: For the `Authorization` header, there's what seems to be a fallback value (an empty string).
Is it necessary?
Especially since for the DPoP header it isn't used.

.github/workflows/test-auth0-api-python.yml
kishore7snehil: The branch-specific trigger ensures the feature branch is  fully tested, while the path filter on PRs prevents unnecessary workflow runs when unrelated files are changed.Anyways, this whole workflow file won't be relevant at all in coming weeks.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+            if len(parts) < 2:
+                raise self._prepare_error(MissingAuthorizationError())
+            elif len(parts) > 2:
+                raise self._prepare_error(
+                    InvalidAuthSchemeError("Invalid Authorization HTTP Header Format for authorization")
+                )
+
+
+        try:
+            scheme, token = authorization_header.split(" ", 1)
```
Widcket, line 97: Given that you've already validated the number of parts above, why not extract the scheme and token from `parts` instead of doing a try/catch here?

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+        except ValueError:
+            raise self._prepare_error(
+                MissingAuthorizationError()
             )
-        return self._jwks_data
+
+
+        scheme = scheme.strip().lower()
+
+        if getattr(self.options, "dpop_required", False) and scheme != "dpop":
```
Widcket, line 106: Since `getattr(self.options, "dpop_required", False)` is used a few times, maybe extract it into an internal helper method?

.github/workflows/test-auth0-api-python.yml
kishore7snehil: Sure, that can be added.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+                if scheme == "bearer":
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError("Invalid scheme. Expected 'DPoP', but got 'bearer'."),
+                        auth_scheme=scheme
+                    )
+                else:
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError("Invalid scheme. Expected 'DPoP' scheme."),
+                        auth_scheme=scheme
```
Widcket, line 115: Could this be simplified by raising an error once and interpolating the `scheme` in the error message (`but got <SCHEME>.`)?

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+                    )
+        if not token.strip():
+            raise self._prepare_error(MissingAuthorizationError())
+
+
+        if scheme == "dpop":
+            if not self.options.dpop_enabled:
```
Widcket, line 122: Could you please elaborate the reasoning for using `getattr(self.options, "dpop_required", False)` in other places and `self.options.dpop_enabled` here?

packages/auth0_api_python/src/auth0_api_python/api_client.py
Widcket: (provided something like https://github.com/auth0/auth0-server-python/pull/22/files#r2244449264 can't just be used instead).

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+                    )
+
+            if "," in dpop_proof:
+                raise self._prepare_error(
+                    InvalidDpopProofError("Multiple DPoP proofs are not allowed"),
+                    auth_scheme=scheme
+                )
+
+            try:
+                await get_unverified_header(dpop_proof)
```
Widcket, line 144: `get_unverified_header` does not seem to have an `await` in it. Does it need to be `async`?

Widcket: Also, given that `get_unverified_header` is called again a few lines below to actually use the returned header value, is it necessary to call it here?

packages/auth0_api_python/src/auth0_api_python/api_client.py
kishore7snehil: We intentionally treat these two flags differently to keep our security posture as clear and robust as possible:

`self.options.dpop_enabled` uses direct access because DPoP support is a fundamental capability that must be explicitly enabled or disabled. When disabled, the entire DPoP authentication flow is unavailable, making it a critical security boundary that should fail fast if misconfigured.

`getattr(self.options, "dpop_required", False)` uses defensive programming for an enforcement policy that determines whether DPoP is mandatory. We allow the SDK to default to a safe, “not required” state if that setting isn’t explicitly provided.

packages/auth0_api_python/src/auth0_api_python/api_client.py
kishore7snehil: This can be accomodated.

packages/auth0_api_python/src/auth0_api_python/api_client.py
kishore7snehil: >Since getattr(self.options, "dpop_required", False) is used a few times, maybe extract it into an internal helper method?

The helper can be added but I don't see too many instances of this.

packages/auth0_api_python/src/auth0_api_python/api_client.py
kishore7snehil: >Also, given that get_unverified_header is called again a few lines below to actually use the returned header value, is it necessary to call it here?

Solving a different purpose at the both the places. The first instance check for malformed `dpop_proof` while the other one is for checking the binding.

Widcket: > Since get_unverified_headeris already defined as an existing async function in utils.py, we must use await when calling it.

Is it a publicly exposed function? If not, it can be fixed.

packages/auth0_api_python/src/auth0_api_python/api_client.py
Widcket: Nevermind, I just saw your comment on why you're using two different was if accessing the options.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+
+
+        if scheme == "dpop":
+            if not self.options.dpop_enabled:
+                raise self._prepare_error(MissingAuthorizationError())
+
+            if not dpop_proof:
+                if getattr(self.options, "dpop_required", False):
+                    raise self._prepare_error(
+                        InvalidAuthSchemeError("Expecting Authorization header with DPoP scheme."),
```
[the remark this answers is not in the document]
Widcket: Then the error message is correct, but the error type is `InvalidAuthSchemeError`.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+                        InvalidAuthSchemeError(
+                            "Operation indicated DPoP use but the request's Authorization HTTP Header scheme is not DPoP"
+                        ),
+                        auth_scheme=scheme
+                    )
+
+            try:
+                claims = await self.verify_access_token(token)
+                if claims.get("cnf") and claims["cnf"].get("jkt"):
+                    if self.options.dpop_enabled:
```
Widcket, line 213: If the access token is DPoP-bound, but the scheme used is Bearer, shouldn't an error be thrown regardless of DPoP being enabled or not?

See requirement:
> When DPoP is enabled in any mode, the SDK must reject any request that includes a DPoP-bound access token in the Authorization header with an authentication scheme other than DPoP. See [Section 7.2](https://www.rfc-editor.org/rfc/rfc9449#section-7.2) of the DPoP specification for more information. The SDK must reject the request with a 400 Bad Request response including an invalid_request error, as described in [Section 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) of the OAuth 2.0 specification.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,210 @@ class ApiClient:
...
+                    access_token=token,
+                    proof=dpop_proof,
+                    http_method=http_method,
+                    http_url=http_url
+                )
+            except InvalidDpopProofError as e:
+                raise self._prepare_error(e, auth_scheme=scheme)
+
+            # DPoP binding verification
+            jwk_dict = (await get_unverified_header(dpop_proof))["jwk"]
```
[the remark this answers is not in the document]
Widcket: Nevermind, I see that this is being done inside `verify_dpop_proof`.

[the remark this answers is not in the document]
kishore7snehil: Yes ,the  verify_dpop_proof is being called before this which has all the `jwk` [checks](https://github.com/auth0/auth0-server-python/pull/22/files#diff-5ca15bae517a69363bf10cc235259709a96b5808fadd39fcb205e7a5e4e9f233R338-R355).

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -120,9 +297,252 @@ async def verify_access_token(
...
+            raise InvalidDpopProofError("jti claim must not be empty")
+
+
+        now = int(time.time())
+        iat = claims["iat"]
+        offset = getattr(self.options, "dpop_iat_offset", 300)    # default 5 minutes
+        leeway = getattr(self.options, "dpop_iat_leeway", 30)     # default 30 seconds
+
+        if not isinstance(iat, int):
+            raise InvalidDpopProofError("Invalid iat claim (must be integer)")
```
Widcket, line 375: The `iat` claim can also be a float.

As per RFC 7519:

[Section 4.1.6](https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.6):
>    The "iat" (issued at) claim identifies the time at which the JWT was
   issued.  This claim can be used to determine the age of the JWT.  Its
   value MUST be a number containing a NumericDate value.  Use of this
   claim is OPTIONAL.


[Section 2](https://www.rfc-editor.org/rfc/rfc7519.html#section-2):
> NumericDate
      A JSON numeric value representing the number of seconds from
      1970-01-01T00:00:00Z UTC until the specified UTC date/time,
      ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1,
      2013 Edition [[POSIX.1](https://www.rfc-editor.org/rfc/rfc7519.html#ref-POSIX.1)] definition "Seconds Since the Epoch", in
      which each day is accounted for by exactly 86400 seconds, other
      than that non-integer values can be represented.  See [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339)
      [[RFC3339](https://www.rfc-editor.org/rfc/rfc3339)] for details regarding date/times in general and UTC in
      particular.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -120,9 +297,252 @@ async def verify_access_token(
...
+
+        if iat < now - offset or iat > now + leeway:
+            raise InvalidDpopProofError("DPoP Proof iat is not recent enough")
+
+        if claims["htm"] != http_method:
```
Widcket, line 380: Should this comparison be case-insensitive?

packages/auth0_api_python/src/auth0_api_python/api_client.py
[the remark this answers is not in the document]
kishore7snehil: IMO the `InvalidAuthSchemeError` class name is appropriate because this is indeed an error with the authentication scheme - the client used the wrong scheme (or no scheme) when DPoP was required.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -120,9 +297,252 @@ async def verify_access_token(
...
+        Args:
+            error: The error to prepare
+            auth_scheme: The authentication scheme that was used ("bearer" or "dpop")
+        """
+        error_code = error.get_error_code()
+        error_description = error.get_error_description()
+
+        www_auth_headers = self._build_www_authenticate(
+            error_code=error_code if error_code != "unauthorized" else None,
+            error_description=error_description if error_code != "unauthorized" else None,
```
Widcket, line 455: This is being checked in two different places; here and inside `_build_www_authenticate`.

Widcket: I mean, `if error_code != "unauthorized"`.

packages/auth0_api_python/README.md
```diff
@@ -6,6 +6,24 @@ It’s intended as a foundation for building more framework-specific integration
...
 📚 [Documentation](#documentation) - 🚀 [Getting Started](#getting-started) - 💬 [Feedback](#feedback)
 
+## Features & Authentication Schemes
+
+This SDK provides comprehensive support for securing APIs with Auth0-issued access tokens:
+
+### **Authentication Schemes**
+- **Bearer Token Authentication** - Traditional OAuth 2.0 Bearer tokens (RS256)
+- **DPoP Authentication** - Enhanced security with Demonstrating Proof-of-Possession (ES256)
+- **Mixed Mode Support** - Seamlessly handle both Bearer and DPoP in the same API
```
Widcket, line 16:
```suggestion
- **Mixed Mode Support** - Seamlessly handles both Bearer and DPoP in the same API
```

packages/auth0_api_python/src/auth0_api_python/utils.py
```diff
@@ -85,4 +90,58 @@ def remove_bytes_prefix(s: str) -> str:
...
+    kty = jwk.get("kty")
+
+    if kty == "EC":
+        if not all(k in jwk for k in ["crv", "x", "y"]):
+            raise ValueError("EC key missing required parameters")
+        members = ("crv", "kty", "x", "y")
+    else:
+        raise ValueError(f"{kty}(Key Type) Parameter missing or unsupported ")
+
+    ordered = {k: jwk[k] for k in members if k in jwk}
```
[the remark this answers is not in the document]
kishore7snehil: In security-critical code like this, that little bit of extra redundancy helps ensure correctness and future-proofs our logic against any subtle changes down the line

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -1,12 +1,28 @@
...
 from pytest_httpx import HTTPXMock
-from unittest.mock import AsyncMock, patch
-
-from src.auth0_api_python.api_client import ApiClient
-from src.auth0_api_python.config import ApiClientOptions
-from src.auth0_api_python.errors import MissingRequiredArgumentError, VerifyAccessTokenError
-from src.auth0_api_python.token_utils import generate_token
 
+# Create public RSA JWK by excluding private key components
+PUBLIC_RSA_JWK = {k: v for k, v in PRIVATE_JWK.items() if k not in ["d", "p", "q", "dp", "dq", "qi"]}
```
Widcket, line 25: Would it to be preferrable to list the allowed property names instead of the other way around?

packages/auth0_api_python/src/auth0_api_python/token_utils.py
```diff
@@ -81,4 +84,136 @@ async def generate_token(
...
+            domain="auth0.local",
+            user_id="user123",
+            audience="my-api",
+            jkt_thumbprint="custom_thumbprint"
+        )
+    """
+
+
+    if jkt_thumbprint is None:
+        public_jwk = {k: v for k, v in PRIVATE_EC_JWK.items() if k != "d"}
```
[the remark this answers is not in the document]
kishore7snehil: The separate filtering operations serve distinct security purposes: 
in `token_utils.py`, filtering explicitly removes private key material (security-focused), while in 
`calculate_jwk_thumbprint`, it ensures only specified RFC 7638 fields are included (compliance-focused).
Combining these operations would obscure their distinct purposes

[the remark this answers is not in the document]
Widcket: The end result is the same.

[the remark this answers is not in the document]
Widcket: Either by including or excluding.

packages/auth0_api_python/tests/test_api_client.py
kishore7snehil: I prefer our current approach of excluding specific private key components rather than listing allowed ones. It makes the security intent crystal clear - we're explicitly removing sensitive data like "d", "p", and "q".

Widcket: This is a test helper. Also, if you're using a blocklist, of there are more fields down the road, you'll need to update the list. Not the case with the allowlist.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+
+    with pytest.raises(MissingRequiredArgumentError) as err:
+        await api_client.verify_dpop_proof(
+            access_token=access_token,
+            proof=dpop_proof,
+            http_method="",  # Empty method
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "http_method" in str(err.value).lower() or "http_url" in str(err.value).lower()
```
Widcket, line 523: This should only assert `http_method`. `http_url` should be asserted in a separate test case.

packages/auth0_api_python/tests/test_api_client.py
kishore7snehil: I agree that having an allow list will be future-proof. but the current approach aligns with how RSA standards define public keys (by absence of private components).

Widcket: Again, this is a *test helper*. No need to overcomplicate things.

kishore7snehil: If we are ready to do a trade-off with the standards, then I can adapt to allowed list approach.

Widcket: That's fine for a test fixture.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+async def test_verify_dpop_proof_fail_invalid_jwk_format():
+    """
+    Test that a DPoP proof with invalid 'jwk' format fails verification.
+    """
+    access_token = "test_token"
+    dpop_proof = await generate_dpop_proof(
+        access_token=access_token,
+        http_method="GET",
+        http_url="https://api.example.com/resource",
+        header_overrides={"jwk": "invalid_jwk"}  # Invalid jwk format
```
Widcket, line 656: This seems to be testing the same thing as the test below. The invalid JWK used here should be invalid for a different reason than having a `d` parameter.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+        http_method="GET",
+        http_url="https://api.example.com/resource"
+    )
+
+    assert result is not None
+
+@pytest.mark.asyncio
+async def test_verify_dpop_proof_iat_past_offset_boundary():
+    """
+    Test IAT validation with timestamps too far in the past.
```
Widcket, line 797:
```suggestion
    Test IAT validation with a timestamp in the past.
```

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "missing required claim" in str(err.value).lower()
+
+@pytest.mark.asyncio
+async def test_verify_dpop_proof_fail_invalid_iat_timing():
+    """
+    Test that a DPoP proof with invalid 'iat' timing fails verification.
```
Widcket, line 736:
```suggestion
        Test IAT validation with a timestamp in the future.
```

packages/auth0_api_python/src/auth0_api_python/api_client.py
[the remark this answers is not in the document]
kishore7snehil: `InvalidDpopProofError` means it will have an error code as `invalid_dpop_proof`. But that's not the case here, whole auth header  is missing here.

packages/auth0_api_python/tests/test_api_client.py
Widcket: My bad, I was confused by `invalid_jwk` having the same name.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "iat is not recent enough" in str(err.value).lower()
+
+@pytest.mark.asyncio
+async def test_verify_dpop_proof_iat_clock_skew_scenarios():
+    """
+    Test IAT validation with various clock skew scenarios.
```
Widcket, line 826: Only one clock skew scenario is being tested. Are any test cases missing?

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "jti claim must not be empty" in str(err.value).lower()
+
+@pytest.mark.asyncio
+async def test_verify_dpop_proof_jti_uniqueness_scenarios():
+    """
+    Test JTI uniqueness and replay protection scenarios.
```
Widcket, line 884: "uniqueness and replay protection scenarios" is not what is being tested here. Just that the `jti` present in the proof is among the returned claims. Which seems a lot like the very first test case in the file.

kishore7snehil: This has been already renamed to `test_verify_dpop_proof_custom_jti_value()`

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+    )
+
+    with pytest.raises(InvalidAuthSchemeError) as err:
+        await api_client.verify_request(
+            headers={"authorization": f"Bearer {token}"},
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "dpop" in str(err.value).lower() or "bearer" in str(err.value).lower()
```
Widcket, line 1370: It's not clear the assertion is verifying that the Bearer token is being rejected.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+    )
+
+    with pytest.raises(InvalidAuthSchemeError) as err:
+        await api_client.verify_request(
+            headers={"authorization": f"Bearer {token}"},
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "cnf" in str(err.value).lower() or "dpop" in str(err.value).lower()
```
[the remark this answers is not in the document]
kishore7snehil: Fixed to match the subset of exact error message.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+    api_client = ApiClient(
+        ApiClientOptions(domain="auth0.local", audience="my-audience")
+    )
+
+    with pytest.raises(MissingAuthorizationError):
+        await api_client.verify_request(
+            headers={},
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
```
Widcket, line 1461: Is this test missing an assertion or does `pytest.raises(MissingAuthorizationError)` check that the specified error is thrown? Because the previous unit test has both `pytest.raises(MissingAuthorizationError)` and a type assertion.

Either the previous test has a superfluous assert, or this one is missing an assert.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+    api_client = ApiClient(
+        ApiClientOptions(domain="auth0.local", audience="my-audience")
+    )
+
+    with pytest.raises(MissingAuthorizationError):
+        await api_client.verify_request(
+            headers={"authorization": "InvalidFormat"},  # Missing scheme and token
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
```
Widcket, line 1478: Same as https://github.com/auth0/auth0-server-python/pull/22/files/c9014aa7c0fd0f2e34ed9a39185c8dcfbbc1da92#r2245826171

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+    api_client = ApiClient(
+        ApiClientOptions(domain="auth0.local", audience="my-audience")
+    )
+
+    with pytest.raises(MissingAuthorizationError):
+        await api_client.verify_request(
+            headers={"authorization": "Basic dXNlcjpwYXNz"},
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
```
Widcket: Same as https://github.com/auth0/auth0-server-python/pull/22/files/c9014aa7c0fd0f2e34ed9a39185c8dcfbbc1da92#r2245826171

packages/auth0_api_python/tests/test_api_client.py
kishore7snehil: This isn't there. Has been removed.

packages/auth0_api_python/tests/test_api_client.py
kishore7snehil: Can't get the reference link.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+
+@pytest.mark.asyncio
+async def test_verify_request_fail_missing_dpop_header():
+    """
+    Test that DPoP scheme requests without DPoP header are rejected.
+    """
+    access_token = "test_token"
+
+    api_client = ApiClient(
+        ApiClientOptions(domain="auth0.local", audience="my-audience")
```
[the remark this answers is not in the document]
kishore7snehil: Yes, because by default we are in allowed mode.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+    )
+    with pytest.raises(InvalidAuthSchemeError) as err:
+        await api_client.verify_request({"authorization": "Bearer  token  with  extra  spaces"})
+    assert "authorization" in str(err.value).lower()
+
+@pytest.mark.asyncio
+async def test_verify_request_with_mixed_case_authorization_header():
+    """Test verify_request with mixed case authorization header."""
+    api_client = ApiClient(ApiClientOptions(domain="auth0.local", audience="my-audience"))
+    with pytest.raises(MissingAuthorizationError):
```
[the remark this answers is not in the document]
kishore7snehil: The link isn't accessible.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+
+    api_client = ApiClient(ApiClientOptions(domain="auth0.local", audience="my-audience"))
+    with pytest.raises(InvalidDpopProofError) as err:
+        await api_client.verify_dpop_proof(
+            access_token=access_token,
+            proof=dpop_proof,
+            http_method="GET",
+            http_url="https://api.example.com/resource"
+        )
+    assert "signature verification failed" in str(err.value).lower()
```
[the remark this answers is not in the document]
kishore7snehil: Gotcha, changed the logic.

packages/auth0_api_python/tests/test_api_client.py
Widcket: This one:

> Is this test missing an assertion or does pytest.raises(MissingAuthorizationError) check that the specified error is thrown? Because the previous unit test has both pytest.raises(MissingAuthorizationError) and a type assertion.
Either the previous test has a superfluous assert, or this one is missing an assert.

packages/auth0_api_python/tests/test_api_client.py
Widcket: This one:

> Is this test missing an assertion or does pytest.raises(MissingAuthorizationError) check that the specified error is thrown? Because the previous unit test has both pytest.raises(MissingAuthorizationError) and a type assertion.
Either the previous test has a superfluous assert, or this one is missing an assert.

packages/auth0_api_python/src/auth0_api_python/utils.py
```diff
@@ -54,20 +54,20 @@ async def fetch_jwks(
             return resp.json()
 
 
-async def get_unverified_header(token: Union[str, bytes]) -> dict:
+def get_unverified_header(token: Union[str, bytes]) -> dict:
     """
     Parse the first segment (header) of a JWT without verifying signature.
-    Ensures correct Base64 padding before decode to avoid garbage bytes.
+    Ensures correct Base64 padding before decode to avoid garbage bytes.\
```
Widcket: There's an added `\` at the end of this line.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -92,28 +96,17 @@ async def verify_request(
...
-                    )
-                else:
-                    raise self._prepare_error(
-                        InvalidAuthSchemeError("Invalid scheme. Expected 'DPoP' scheme."),
-                        auth_scheme=scheme
-                    )
+        if self.is_dpop_required() and scheme != "dpop":
+                raise self._prepare_error(
+                    InvalidAuthSchemeError(
+                        f"Invalid scheme. Expected DPoP{', but got ' + scheme + '.' if scheme and scheme != 'dpop' else ' scheme.'}"
```
Widcket, line 106: The `if` in line 103 already checks that scheme is not `dpop`. Why is the check `if scheme and scheme != 'dpop'` needed?

packages/auth0_api_python/tests/test_api_client.py
Widcket: I see it was changed from a type assertion into a status code assertion:
Screenshot 2025-07-31 at 18 41 45

kishore7snehil: Yes, I felt that would make more sense.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +404,1153 @@ async def test_verify_access_token_fail_no_audience_config():
...
+            http_url="https://api.example.com/resource"
+        )
+
+    assert "multiple" in str(err.value).lower()
+
+@pytest.mark.asyncio
+async def test_verify_request_with_empty_token():
+    """Test verify_request with empty token value."""
+    api_client = ApiClient(ApiClientOptions(domain="auth0.local", audience="my-audience"))
+    with pytest.raises(MissingAuthorizationError):
```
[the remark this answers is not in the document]
kishore7snehil: This is already fixed pointing to status code assertion.

.github/workflows/test-auth0-api-python.yml
kishore7snehil: It's done.

packages/auth0_api_python/src/auth0_api_python/api_client.py
kishore7snehil: The logic has been updated now.

packages/auth0_api_python/EXAMPLES.md
```diff
@@ -0,0 +1,160 @@
...
+
+# Example usage
+headers = {"authorization": "Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..."}
+result = asyncio.run(validate_request(headers))
+```
+
+
+## DPoP Authentication 
+
+**DPoP (Demonstrating Proof-of-Possession)** is a security extension that binds access tokens to cryptographic keys, preventing token theft and replay attacks.
```
Widcket, line 70:
```suggestion
**DPoP (Demonstrating Proof-of-Possession)** is a security extension that binds access tokens to cryptographic keys, preventing the replay of stolen or leaked access tokens.
```

Widcket: It doesn't prevent token theft.

Widcket: That means it protects against the consequences of token theft (using them to access protected resources), but it doesn't prevent the token theft itself.

Widcket:
How is the library handling the case when the Authorization header contains a malformed access token (e.g. `"Authorization: Bearer invalid-format"`)? It should be an invalid token error (status 401, `invalid_token` code).

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -120,9 +291,254 @@ async def verify_access_token(
...
+        Args:
+            error: The error to prepare
+            auth_scheme: The authentication scheme that was used ("bearer" or "dpop")
+        """
+        error_code = error.get_error_code()
+        error_description = error.get_error_description()
+
+        www_auth_headers = self._build_www_authenticate(
+            error_code=error_code,
+            error_description=error_description,
```
Widcket, line 451: Should the error description here have any double quotes (`"`) escaped (e.g. by replacing with single quotes)?

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -120,9 +291,254 @@ async def verify_access_token(
...
+
+        if iat < now - offset:
+            raise InvalidDpopProofError("DPoP Proof iat is too old")
+        elif iat > now + leeway:
+            raise InvalidDpopProofError("DPoP Proof iat is from the future")
+
+        if claims["htm"].lower() != http_method.lower():
+            raise InvalidDpopProofError("DPoP Proof htm mismatch")
+
+        if normalize_url_for_htu(claims["htu"]) != normalize_url_for_htu(http_url):
```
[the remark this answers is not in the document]
Widcket: It should be an invalid proof response.

packages/auth0_api_python/tests/test_api_client.py
```diff
@@ -388,3 +405,1137 @@ async def test_verify_access_token_fail_no_audience_config():
...
+        access_token=access_token,
+        proof=dpop_proof,
+        http_method="GET",
+        http_url="https://api.example.com/resource"  # Lowercase domain
+    )
+    assert result is not None
+
+
+@pytest.mark.asyncio
+async def  test_verify_dpop_proof_htu_trailing_slash_mismatch():
```
Widcket, line 1074: Please add another test case showing that when the query params and fragments are removed, the trailing slash is preserved.

E.g:
```
got = 'https://api.example.com/resource/?abc=def'
expected = 'https://api.example.com/resource/'
```

Widcket:
How is the library handling the case when there the DPoP proof contains extra claims in the payload, and extra properties in the header? These should be ignored.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -14,46 +29,204 @@ class ApiClient:
...
+
+
+        parts = authorization_header.split(" ")
+        if len(parts) != 2:
+            if len(parts) < 2:
+                raise self._prepare_error(MissingAuthorizationError())
+            elif len(parts) > 2:
+                raise self._prepare_error(
+                    InvalidAuthSchemeError("Invalid Authorization HTTP Header Format for authorization")
+                )
```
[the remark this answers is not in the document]
kishore7snehil: Yes, and it's behaving in the similar manner (**for DPoP required mode**).

**For Required Mode:**

Expectation: 
image

Result: 
image

**For Allowed  Mode:**

Expectation: 
image

Result: 
image



**For Neither of the Mode **

Expectation: 
image

Result: 
image

kishore7snehil:
> How is the library handling the case when the Authorization header contains a malformed access token (e.g. `"Authorization: Bearer invalid-format"`)? It should be an invalid token error (status 401, `invalid_token` code).

Please check the following out :


**For Neither of the Mode **

- Bearer Request 

 image
 
 - DPoP Request
 
image


**For Allowed  Mode:**

- Bearer Request 

image
 
 - DPoP Request
 
image

**For Required Mode:**

- Bearer Request 

image
 
 - DPoP Request
 
image

Widcket:
Thanks. Seems that there is an instance where the `WWW-Authenticate` header is just `Bearer`, and is missing the `error` and `error_description`.

packages/auth0_api_python/src/auth0_api_python/api_client.py
kishore7snehil: Based on the integration test output, there's no issue with the current quoting. The current code is working correctly, so no changes are needed.
image

packages/auth0_api_python/src/auth0_api_python/api_client.py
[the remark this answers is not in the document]
kishore7snehil: This is working as expected now.

packages/auth0_api_python/src/auth0_api_python/api_client.py
```diff
@@ -120,9 +291,259 @@ async def verify_access_token(
...
+            raise MissingRequiredArgumentError("http_method/http_url")
+
+        header = get_unverified_header(proof)
+
+        if header.get("typ") != "dpop+jwt":
+            raise InvalidDpopProofError("Unexpected JWT 'typ' header parameter value")
+
+        alg = header.get("alg")
+        if alg not in self._dpop_algorithms:
+            raise InvalidDpopProofError(f"Unsupported alg: {alg}")
```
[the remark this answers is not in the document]
Widcket: Also for all other error messages where this applies.

packages/auth0_api_python/src/auth0_api_python/api_client.py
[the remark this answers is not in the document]
Widcket: Also, please note that the case I raised on this thread is not `random_string valid_token" but just `random_string`. That should always result in a 400 IIRC.

packages/auth0_api_python/src/auth0_api_python/api_client.py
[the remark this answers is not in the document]
kishore7snehil: This is taken care.

PR #103, opened 2026-04-21
feat: add Auth0-Client telemetry header

kishore7snehil:
### Changes

- Add `Telemetry` class (`telemetry.py`) that builds and caches the `Auth0-Client` header - a base64-encoded JSON payload with SDK name, version, and Python runtime environment
- Add `_get_http_client()` helper to `ServerClient`, `MyAccountClient`, and `MfaClient` to inject telemetry headers into all `httpx` requests (17 call sites)
- Pass telemetry headers to `AsyncOAuth2Client` for token exchange calls
- Add unit tests for telemetry header format and integration

### DIFF
--- a/src/auth0_server_python/telemetry.py
+++ b/src/auth0_server_python/telemetry.py
@@ -44,6 +44,6 @@ class Telemetry:
         """Create a Telemetry instance with this SDK's package metadata."""
         try:
             version = importlib.metadata.version(Telemetry._PACKAGE_NAME)
-        except Exception:
+        except importlib.metadata.PackageNotFoundError:
             version = "unknown"
         return Telemetry(name=Telemetry._PACKAGE_NAME, version=version)
--- a/src/auth0_server_python/telemetry.py
+++ b/src/auth0_server_python/telemetry.py
@@ -5,12 +5,11 @@ Builds and caches the Auth0-Client and User-Agent headers sent
 on every HTTP request to Auth0 endpoints.
 """
 
-from __future__ import annotations
-
 import base64
 import importlib.metadata
 import json
 import platform
+from typing import Optional
 
 
 class Telemetry:
@@ -18,11 +17,11 @@ class Telemetry:
 
     _PACKAGE_NAME = "auth0-server-python"
 
-    def __init__(self, name: str, version: str, env: dict[str, str] | None = None):
+    def __init__(self, name: str, version: str, env: Optional[dict[str, str]] = None):
         self.name = name
         self.version = version
         self.env = env if env is not None else {"python": platform.python_version()}
-        self._cached_headers: dict[str, str] | None = None
+        self._cached_headers: Optional[dict[str, str]] = None
 
     def get_headers(self) -> dict[str, str]:
         """Return the telemetry headers, building and caching on first call."""
@@ -41,7 +40,7 @@ class Telemetry:
         return self._cached_headers
 
     @staticmethod
-    def default() -> Telemetry:
+    def default() -> "Telemetry":
         """Create a Telemetry instance with this SDK's package metadata."""
         try:
             version = importlib.metadata.version(Telemetry._PACKAGE_NAME)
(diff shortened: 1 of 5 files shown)


arpit-jn (dismissed):
LGTM

src/auth0_server_python/telemetry.py
```diff
@@ -0,0 +1,49 @@
...
+class Telemetry:
+    """Builds and caches telemetry headers for Auth0 HTTP requests."""
+
+    _PACKAGE_NAME = "auth0-server-python"
+
+    def __init__(self, name: str, version: str, env: Optional[dict[str, str]] = None):
+        self.name = name
+        self.version = version
+        self.env = env if env is not None else {"python": platform.python_version()}
+        self._cached_headers: Optional[dict[str, str]] = None
```
nandan-bhat: The headers are built lazily here, but `get_headers()` is only ever called once in `ServerClient.__init__`, where the result is immediately stored in `_telemetry_headers`. The cache in `Telemetry` is never needed.

You can directly assign it like this and use in  `_telemetry_headers`
```
  def __init__(self, name, version, env=None):
      self.name = name                                                                                                                              
      self.version = version                                                                                                                        
      self.env = env if env is not None else {"python": platform.python_version()}                                                                  
      payload = {"name": self.name, "version": self.version, "env": self.env}                                                                       
      self.headers = {                   # ← built once, right here                                                                                 
          "Auth0-Client": base64.b64encode(json.dumps(payload).encode()).decode(),                                                                  
          "User-Agent": f"Python/{platform.python_version()}",                                                                                      
      }  
```

src/auth0_server_python/tests/test_telemetry.py
```diff
@@ -0,0 +1,142 @@
...
+        mock_response.status_code = 200
+        mock_response.json.return_value = {"issuer": "https://auth0.local/"}
+        mock_response.raise_for_status = MagicMock()
+
+        mock_http_client = AsyncMock()
+        mock_http_client.get.return_value = mock_response
+        mock_http_client.__aenter__ = AsyncMock(return_value=mock_http_client)
+        mock_http_client.__aexit__ = AsyncMock(return_value=False)
+
+        mocker.patch.object(client, "_get_http_client", return_value=mock_http_client)
```
nandan-bhat: This test patches `_get_http_client` and asserts it was called once but it doesn't verify the headers on the `HTTP` request. The test name is `test_fetch_oidc_metadata_sends_telemetry`, but it only checks that `_get_http_client` was invoked, not that telemetry headers were actually present.

Can we assert the header values here to match the test name ?

src/auth0_server_python/tests/test_telemetry.py
```diff
@@ -0,0 +1,142 @@
...
+    @patch(
+        "auth0_server_python.telemetry.importlib.metadata.version",
+        side_effect=importlib.metadata.PackageNotFoundError("not installed"),
+    )
+    def test_default_factory_unknown_version_on_error(self, _mock):
+        telemetry = Telemetry.default()
+        assert telemetry.version == "unknown"
+
+
+class TestServerClientTelemetry:
```
nandan-bhat, line 64: There are good tests for the `_get_http_client` path, but no test covering `complete_interactive_login` which is the only place
`self._oauth.fetch_token(...)` is called. That code path goes through `Authlib`'s `AsyncOAuth2Client`, not through `_get_http_client`, so it's a different code path entirely.

Can we add a test that verifies telemetry headers are sent during the authorization code exchange ?

kishore7snehil: Did an integration testing. Everything looks good

src/auth0_server_python/tests/test_telemetry.py
kishore7snehil: Fixed in [8173621](https://github.com/auth0/auth0-server-python/pull/103/changes/81736216fd2b3b12d75aa49c769e37a527c5adba)

src/auth0_server_python/telemetry.py
kishore7snehil: Addressed in [8173621](https://github.com/auth0/auth0-server-python/pull/103/changes/81736216fd2b3b12d75aa49c769e37a527c5adba)

nandan-bhat (approved):

PR #114, opened 2026-05-26
feat: Organisations support

rmad17:
### Changes
  - Org login support: organization on ServerClient (client-level default) and StartInteractiveLoginOptions (per-login override); org_id / org_name claim validation enforced automatically at callback
  - Org invitation flow: invitation as a first-class typed field on StartInteractiveLoginOptions; forwarded to /authorize alongside organization
  - Org token validation error: OrganizationTokenValidationError raised when the org_id or org_name claim in the returned token does not match what was requested at login; all other org errors from Auth0 (access
  denied, invalid request, invitation rejected) surface as ApiError with the raw OAuth error code and error_description

###Out of Scope
  - Org enforcement on backchannel_authentication_grant, login_with_custom_token_exchange, and get_access_token refresh-path token exchange
### References
https://auth0team.atlassian.net/browse/SDK-8833

### Testing

Please describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors.

- [ ] This change adds integration test coverage

### Checklist


src/auth0_server_python/auth_server/server_client.py
```diff
@@ -69,7 +74,50 @@
...
+)
+
+_INVITATION_DESCRIPTIONS = (
+    "invalid_user_invitation_ticket",
+    "invitation not found or already used",
+    "invalid invitation ticket",
+)
+
+
+def _classify_org_error(error: str, error_description: str) -> "ApiError":
```
kishore7snehil: We're branching on the text of `error_description`- that's just a human-readable message, so if the server rewords it, this quietly falls through to a generic `ApiError` and any caller depending on the typed error breaks, with nothing flagging it. We should drop the classifier and just raise `ApiError(error, error_description)` like the rest of the callback already does

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -69,7 +74,50 @@
 # redirect_uri is intentionally excluded — in MCD mode it is built
 # dynamically from the resolved domain at login time.
 INTERNAL_AUTHORIZE_PARAMS = ["client_id", "response_type",
-                             "code_challenge", "code_challenge_method", "state", "nonce", "scope"]
+                             "code_challenge", "code_challenge_method", "state", "nonce", "scope",
+                             "organization"]
+
+_ORG_ACCESS_DENIED_DESCRIPTIONS = (
```
kishore7snehil: This is feeding the string-matching I flagged on `_classify_org_error` - can come out along with it.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -69,7 +74,50 @@
...
+    "is not part of the",
+    "quota exceeded",
+    "organization member limit",
+    "user_id that longer than 1024",
+)
+
+_ORG_INVALID_REQUEST_DESCRIPTIONS = (
```
kishore7snehil: Same as [above](https://github.com/auth0/auth0-server-python/pull/114/changes#r3434337264).

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -69,7 +74,50 @@
...
+    "organizations feature is not supported",
+    "parameter organization is required",
+    "parameter organization is not allowed",
+    "parameter organization is invalid",
+    "organization name must have between",
+    "organization must only contain lowercase",
+    "client is missing",
+)
+
+_INVITATION_DESCRIPTIONS = (
```
kishore7snehil: Same as [above](https://github.com/auth0/auth0-server-python/pull/114/changes#r3434337264).

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -603,6 +699,8 @@ async def complete_interactive_login(
         if "error" in query_params:
             error = query_params.get("error", [""])[0]
             error_description = query_params.get("error_description", [""])[0]
+            if transaction_data.organization:
```
kishore7snehil: We can lose this branch entirely and just let it hit the plain `ApiError` raise right below it.

src/auth0_server_python/error/__init__.py
```diff
@@ -185,6 +185,59 @@ def __init__(self, message: str):
...
+    against the organization value that was requested at login.
+    """
+    code = "organization_token_validation_error"
+
+    def __init__(self, message: str):
+        super().__init__(message)
+        self.name = "OrganizationTokenValidationError"
+
+
+class OrganizationRequiredError(Auth0Error):
```
kishore7snehil: This one only exists to back the string-matcher we're removing, so I'd drop it. I think `OrganizationTokenValidationError` is the only org error we actually need.

src/auth0_server_python/error/__init__.py
```diff
@@ -185,6 +185,59 @@ def __init__(self, message: str):
...
+    """
+    code = "organization_required_error"
+
+    def __init__(self, message: str, cause: Optional[Exception] = None):
+        super().__init__(message)
+        self.name = "OrganizationRequiredError"
+        self.cause = cause
+
+
+class OrganizationAccessDeniedError(Auth0Error):
```
kishore7snehil: Same as [above](https://github.com/auth0/auth0-server-python/pull/114/changes#r3434337291)

src/auth0_server_python/error/__init__.py
```diff
@@ -185,6 +185,59 @@ def __init__(self, message: str):
...
+    """
+    code = "organization_access_denied_error"
+
+    def __init__(self, message: str, cause: Optional[Exception] = None):
+        super().__init__(message)
+        self.name = "OrganizationAccessDeniedError"
+        self.cause = cause
+
+
+class OrganizationInvitationError(Auth0Error):
```
kishore7snehil: Same as [above](https://github.com/auth0/auth0-server-python/pull/114/changes#r3434337291)

src/auth0_server_python/error/__init__.py
```diff
@@ -185,6 +185,59 @@ def __init__(self, message: str):
...
+
+class OrganizationRequiredError(Auth0Error):
+    """
+    Raised when Auth0 rejects the authorization request due to an organization
+    configuration problem — invalid format, feature disabled, client not
+    configured for organizations, or the organization does not exist.
+    """
+    code = "organization_required_error"
+
+    def __init__(self, message: str, cause: Optional[Exception] = None):
```
kishore7snehil: All these three take a `cause` argument that nothing ever passes in, so it's just sitting unused.We don't leave a dead param hanging around on anything that does stay.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -69,7 +74,50 @@
 # redirect_uri is intentionally excluded — in MCD mode it is built
 # dynamically from the resolved domain at login time.
 INTERNAL_AUTHORIZE_PARAMS = ["client_id", "response_type",
-                             "code_challenge", "code_challenge_method", "state", "nonce", "scope"]
+                             "code_challenge", "code_challenge_method", "state", "nonce", "scope",
+                             "organization"]
```
kishore7snehil: Adding `organization` here means it can't be passed through `authorization_params` anymore, so the typed field becomes the only way in. We shouldn't do this and also while writing the Organization section in the doc, mention that we can pass it through `authorization_params`.

examples/Organizations.md
```diff
@@ -0,0 +1,193 @@
+# Organizations
```
kishore7snehil: This is a lot of doc for what's really just a couple of login options. I'd fold the essentials into the existing [interactive-login example](https://github.com/auth0/auth0-server-python/blob/main/examples/InteractiveLogin.md) and the README, and link out to the [hosted Auth0 org docs](https://auth0.com/docs/manage-users/organizations) rather than re-documenting the whole feature here.

examples/Organizations.md
```diff
@@ -0,0 +1,193 @@
...
+        options,
+        store_options={"request": request, "response": response}
+    )
+    return RedirectResponse(url=authorization_url)
+```
+
+> [!NOTE]
+> `organization` and `invitation` are forwarded to `/authorize`. Auth0 consumes the invitation ticket server-side — it is not stored in the encrypted transaction. If the ticket is expired or already used, `complete_interactive_login` raises `OrganizationInvitationError`.
+
+## 4. Handling organization errors
```
kishore7snehil: This whole section is walking through the error classes we're taking out - once it's plain `ApiError` there's nothing org-specific left to show, so I think this can just go.

README.md
```diff
@@ -173,6 +173,33 @@ The SDK handles per-domain OIDC discovery, JWKS fetching, issuer validation, and
...
+)
+
+# Multi-org: pass organization per login
+authorization_url = await auth0.start_interactive_login(
+    StartInteractiveLoginOptions(organization="org_xyz789"),
+    store_options={"request": request, "response": response}
+)
+```
+
+The SDK validates the `org_id` (or `org_name`) claim in the returned token automatically — no extra code needed at callback. For invitation flows, per-org error handling, and reading org data from the session, see [examples/Organizations.md](examples/Organizations.md).
```
kishore7snehil: If we drop the standalone file, this link will need to go or point somewhere else.

kishore7snehil (changes requested):

src/auth0_server_python/auth_server/server_client.py
rmad17: ~The idea behind this was to catch some known invalid_request caused by org login related changes like invalid_org_id, etc. and share that information to the developer. It would be better DX. The fallback still remains ApiError.~
Will revert this as this could also become a point of failure if the description changes.

kishore7snehil: Surfacing org failures clearly is good DX but with plain `ApiError(error, error_description)` the developer already gets both the stable error code (`access_denied`, `invalid_request`) and the full description text, so no information is actually lost. What the classifier adds on top is matching against the text of `error_description`, and that's the part we should avoid.
 

> invalid_org_id

There isn't actually a distinct invalid_org_id error code coming back; these arrive as invalid_request / access_denied with the detail only in the free-form message

src/auth0_server_python/error/__init__.py
```diff
@@ -185,6 +185,19 @@ def __init__(self, message: str):
         self.name = "StartLinkUserError"
 
 
+class OrganizationTokenValidationError(Auth0Error):
```
kishore7snehil: Can we move this below `class AccessTokenErrorCode`

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -263,7 +268,6 @@ class CustomTokenExchangeOptions(BaseModel):
     scope: Optional[str] = None
     actor_token: Optional[str] = None
     actor_token_type: Optional[str] = None
-    organization: Optional[str] = None
```
kishore7snehil, line 266: Why did we delete this?

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -319,7 +323,6 @@ class LoginWithCustomTokenExchangeOptions(BaseModel):
...
```
kishore7snehil, line 322: Why did we delete this?

kishore7snehil (changes requested):

examples/InteractiveLogin.md
```diff
@@ -137,4 +137,99 @@ print(result.get("authorization_details"))               # Rich Authorization Re
...
+|---|---|
+| `access_denied` | User not a member, connection not enabled for org, member quota exceeded |
+| `invalid_request` | Invalid org format, feature disabled, client not configured for orgs, expired or invalid invitation ticket |
+
+### Reading organization data from the session
+
+After a successful org login, `org_id` and `org_name` are available on the user object:
+
+```python
+user = await auth0.get_user(store_options={"request": request, "response": response})
```
kishore7snehil, line 231: The [line](https://github.com/auth0/auth0-server-python/pull/114/changes#diff-f42a9c2261e1a49f088ab83b464129f5005763362469abc460ba58c6d207d408R171) here says that we can pass either org_id or org_name but here we are showing both the claims. Does the claim contain both the values even if we pass one?

kishore7snehil (changes requested):

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: @rmad17 You should check this once. If it's intentional, then only we should keep it.

kishore7snehil (changes requested):

examples/InteractiveLogin.md
rmad17: Only if login via org_name is enabled from tenant settings.

src/auth0_server_python/auth_server/server_client.py
rmad17: Initial thought was to intentionally not allow. Though will change the approach as other SDKs also follow a pattern to have an alternate option similar to this.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -207,6 +214,41 @@ def _normalize_url(self, value: str) -> str:
 
         return value.rstrip('/')
 
+    def _validate_org_claims(self, claims: dict, expected_org: str) -> None:
```
kishore7snehil: We should  Move `_validate_org_claims` out of the `ServerClient` class and into `src/auth0_server_python/utils/helpers.py `as a module-level function. Something like this

`def validate_org_claims(claims: dict, expected_org: str) -> None:`

kishore7snehil (changes requested):

kishore7snehil (approved):
LGTM!

PR #116, opened 2026-06-01
feat: passkeys, myaccount with dpop support

rmad17:
## 📋 Changes

This PR adds Passkey Authentication and My Account API Authentication Methods
support to auth0-server-python. Passkey sign-in is a two-step challenge/verify
ceremony that establishes a server-side session. The My Account surface lets a
logged-in user enroll, list, update, and delete their own authentication
methods. Both surfaces support DPoP (RFC 9449) sender-constrained tokens via
an optional parameter.

### Features

- **Passkey sign-in (`ServerClient`):** `passkey_signup_challenge`,
  `passkey_login_challenge`, and `signin_with_passkey`. Call
  `passkey_signup_challenge` or `passkey_login_challenge` to get an
  `auth_session` and WebAuthn options for the browser, then call
  `signin_with_passkey` with the signed credential to complete sign-in and
  create a session.
- **My Account API — Authentication Methods & Factors (`MyAccountClient`):**
  `get_factors`, `list_authentication_methods`, `get_authentication_method`,
  `delete_authentication_method`, `update_authentication_method`,
  `enroll_authentication_method`, `verify_authentication_method`. Enrollment
  and verification are generic across types (passkey, email, phone, TOTP,
  push, recovery-code, password) — the type is set via the request model.
- **DPoP (RFC 9449):** new `DPoPAuth` auth scheme. Every passkey and My
  Account method takes an optional `dpop_key` (EC P-256 key); when supplied,
  the SDK attaches a DPoP proof to the request. Omitting `dpop_key` uses the
  existing Bearer flow.
- **MFA on passkey sign-in and refresh:** `signin_with_passkey` and
  `get_token_by_refresh_token` now raise `MfaRequiredError` when a second
  factor is required, matching `get_access_token`.

### API Changes

- New public methods: `ServerClient.passkey_signup_challenge`,
  `passkey_login_challenge`, `signin_with_passkey`; `MyAccountClient.get_factors`,
  `list_authentication_methods`, `get_authentication_method`,
  `delete_authentication_method`, `update_authentication_method`,
  `enroll_authentication_method`, `verify_authentication_method`.
- New public models: `PasskeyUserProfile`, `PasskeySignupChallengeResponse`,
  `PasskeyLoginChallengeResponse`, `PasskeyAuthResponse`,
  `PasskeyTokenResponse`, `PasskeyLoginResult`, `Factor`, `GetFactorsResponse`,
  `AuthenticationMethod`, `ListAuthenticationMethodsResponse`,
  `EnrollAuthenticationMethodRequest`, `EnrollmentChallengeResponse`,
  `VerifyAuthenticationMethodRequest`, `UpdateAuthenticationMethodRequest`.
- New error types: `PasskeyError`, `PasskeyErrorCode`. Raised by the three
  `ServerClient` passkey methods. `MyAccountClient` methods continue to raise
  `MyAccountApiError` / `ApiError`.
- New export: `auth_schemes.DPoPAuth`.
- New optional `dpop_key` parameter on every passkey and My Account method.
- `AuthenticatorResponse.authenticator_type` and `ChallengeResponse.challenge_type`
  are now typed `str` (previously `Literal`), so an authenticator/challenge type
  Auth0 adds later no longer raises a `ValidationError`. No change to values
  returned today.

### Deprecation

**Deprecation (non-breaking):** the `AuthenticatorType` alias still imports and
resolves to its previous `Literal["otp", "oob", "recovery-code"]`, but now emits
a `DeprecationWarning` on access and is scheduled for removal in a future major
release. It no longer constrains a model field. Existing imports keep working.


src/auth0_server_python/auth_types/__init__.py
```diff
@@ -381,50 +413,252 @@ class ConnectedAccountConnection(BaseModel):
...
+    model_config = ConfigDict(populate_by_name=True)
+    id: str
+    raw_id: str = Field(alias="rawId")
+    type: str
+    authenticator_attachment: Optional[str] = Field(None, alias="authenticatorAttachment")
+    response: dict[str, str]
+    client_extension_results: Optional[dict] = Field(None, alias="clientExtensionResults")
+
+
+class VerifyAuthenticationMethodRequest(BaseModel):
```
kishore7snehil: **Are these patterns necessary, or generated boilerplate?**

A few things on the new models look possibly over-engineered — can you confirm each is intentional?

- **`__repr__` redaction (`[REDACTED]`)** is added to the new passkey models, but existing secret-bearing models in this same file (`TokenSet`, `MfaVerifyResponse` — both hold `access_token` / `refresh_token` / `id_token`) have no such redaction. Either we adopt redaction project-wide or drop it here for consistency; having it only on the new models is inconsistent.
- **`@model_validator` (`_check_at_least_one_method`)** here — is this a real requirement or speculative validation?
- **`model_config = ConfigDict(populate_by_name=True)`** — what's the reason this is needed on these models? If nothing populates by field name (vs alias), it can be removed.
- **`EnrollAuthenticationMethodRequest.type` is a free-form `str`** while this file already uses `Literal[...]` for closed sets (`AuthenticatorType`, `OobChannel`, etc.). `type` (and `preferred_authentication_method`, which the spec restricts to `sms`/`voice`) could be `Literal` so bad values fail at construction instead of as a remote API error.
- Minor: the enroll/verify docstrings say `create:me:authentication_methods` (underscore); the spec scope uses a hyphen — `create:me:authentication-methods`.

src/auth0_server_python/auth_schemes/dpop_auth.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __repr__(self) -> str:
+        return "DPoPAuth(token=[REDACTED], key=[REDACTED])"
+
+    def __str__(self) -> str:
+        return "DPoPAuth(token=[REDACTED], key=[REDACTED])"
+
+    def auth_flow(self, request: httpx.Request):
+        proof = self._make_proof(request.method, str(request.url))
+        request.headers["Authorization"] = f"DPoP {self._token}"
```
kishore7snehil, line 80: Two DPoP gaps:

1. `DPoPAuth` has no `DPoP-Nonce` handling. Auth0/Okta commonly responds `401` + a `DPoP-Nonce` header and expects the proof retried with that nonce, so the first call against a nonce-requiring resource server will fail. Can we track nonce-retry even if it's deferred past GA?
2. Separately, `signin_with_passkey` does a plain token POST with no DPoP proof, and `PasskeyTokenResponse.token_type` defaults to `Bearer`, but the Test Plan says `/oauth/token` (webauthn) returns Bearer *or* DPoP based on tenant setup. If the tenant issues a DPoP-bound token, the SDK ends up holding a token bound to a key it never used. Can we confirm the GA scope for DPoP-bound passkey tokens?

src/auth0_server_python/tests/test_passkey_server_client.py
```diff
@@ -0,0 +1,585 @@
+import time
```
kishore7snehil: Tests for methods that live in `server_client.py` should be added to `test_server_client.py` as a new section that mirrors the existing test patterns in that file (same fixtures, `mocker.patch("httpx.AsyncClient.post", ...)` / `auth=ANY` style). This file uses a different mocking idiom (patching `_get_http_client` with a hand-built async context manager) than the rest of the suite.  let's keep it consistent.

Also note: the connect-account methods in `my_account_client.py` were reshaped alongside the formatting pass - please confirm no behavior changed there.

src/auth0_server_python/tests/test_passkey_my_account.py
```diff
@@ -0,0 +1,830 @@
+from unittest.mock import AsyncMock, MagicMock
```
kishore7snehil: Let's not introduce feature-scoped test files (`test_passkey_my_account.py`, `test_passkey_server_client.py`) at this point. We can keep using the existing `test_my_account_client.py` and `test_server_client.py` until we have a concrete reason to split files. Please fold these tests into the existing files.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
                 e
             )
 
+    # ============================================================================
+    # PASSKEY AUTHENTICATION
+    # ============================================================================
+
+    GRANT_TYPE_PASSKEY = "urn:okta:params:oauth:grant-type:webauthn"
```
kishore7snehil: Please pull these up to a module/class-level constant rather than constructing them in the middle of the method body, consistent with how the rest of the client references endpoints.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
                 e
             )
 
+    # ============================================================================
```
kishore7snehil, line 2638: The new Passkey section is inserted in the middle of `ServerClient` (between Custom Token Exchange and the trailing MFA section), rather than appended after the existing methods. Can we move the new feature block to the end of the class so the diff reads as a pure addition and doesn't visually interleave with existing code? 

The same applies to the new models in `auth_types/__init__.py` , new additions should go after the existing model definitions.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
...
+        email: Optional[str] = None,
+        username: Optional[str] = None,
+        phone_number: Optional[str] = None,
+        given_name: Optional[str] = None,
+        family_name: Optional[str] = None,
+        nickname: Optional[str] = None,
+        picture: Optional[str] = None,
+        user_metadata: Optional[dict[str, Any]] = None,
+        connection: Optional[str] = None,
+        organization: Optional[str] = None,
```
kishore7snehil: This method exposes ~11 flat keyword arguments and assembles a nested `user_profile` dict field-by-field inline.Rather than spreading the fields across kwargs and hand-building/remapping the dict inline, can we introduce a typed request model (e.g. `PasskeySignupChallengeOptions`) with a nested `user_profile` sub-model? 

Using `ConfigDict(extra="allow")` on that sub-model would let new/unanticipated profile fields pass through   without a code change, isolating the schema churn to one place.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
                 e
             )
 
+    # ============================================================================
+    # PASSKEY AUTHENTICATION
+    # ============================================================================
+
+    GRANT_TYPE_PASSKEY = "urn:okta:params:oauth:grant-type:webauthn"
+
```
kishore7snehil, line 2650: These three methods implement the passkey login/signup ceremony: passkey_signup_challenge / `passkey_login_challenge` (step 1: get `auth_session` + the WebAuthn challenge), then the app runs navigator.credentials.create/get, then `signin_with_passkey` (step 2 :exchange the signed credential for tokens via the webauthn grant). Two questions:

1. Can you confirm this is the intended end-to-end flow and that all three methods are exercised by it?
2. `signin_with_passkey` returns a `PasskeyTokenResponse` but, unlike complete_interactive_login, never persists to the state store or creates a session. Is that intentional (caller owns the session), or should passkey sign-in integrate with the SDK's existing session/state handling like every other login path here?

For an RWA SDK whose value-add is session/state management, returning bare tokens reads as an inconsistency.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
...
+            body: dict[str, Any] = {"client_id": self._client_id}
+            if self._client_secret:
+                body["client_secret"] = self._client_secret
+            if user_profile:
+                body["user_profile"] = user_profile
+            if connection:
+                body["realm"] = connection
+            if organization:
+                body["organization"] = organization
+
```
kishore7snehil: These `if <field> is not None: user_profile[...] = <field>` assignments are manually replicating what Pydantic gives for free. If we model `user_profile` as a typed sub-model on a request object (e.g. `PasskeySignupChallengeOptions`), this whole block reduces to:                                                                                                       
```
if options.user_profile:                                                                                                                                                               
      body["user_profile"] = options.user_profile.model_dump(exclude_none=True) 
``` 
`exclude_none=True` preserves the current "omit unset keys" behavior, and each field gets real typing instead of an untyped `Optional[str] `kwarg. 

We can use  `ConfigDict(extra="allow")` to the sub-model also which will allow new profile fields pass through without touching this method.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
...
+            url = f"https://{domain}/passkey/register"
```
kishore7snehil: This can be moved [here](https://github.com/auth0/auth0-server-python/pull/116/changes#diff-f89e94b52d5636391323612572af43429518b8d6aa24a4f9e60e9c8489d3bc7bR2565)

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
...
+
+                if response.status_code != 200:
+                    try:
+                        error_data = response.json()
+                    except (json.JSONDecodeError, ValueError):
+                        raise ApiError(
+                            "passkey_challenge_error",
+                            f"Passkey signup challenge failed with status {response.status_code}",
+                        )
+                    raise ApiError(
```
kishore7snehil: Consider a dedicated `PasskeyError(Auth0Error)` class (mirroring  `CustomTokenExchangeError`) instead of `ApiError` with a `passkey_challenge_error` string code, so callers can catch the type directly.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2475,6 +2479,288 @@ async def login_with_custom_token_exchange(
...
+                        "invalid_response",
+                        "Failed to parse passkey signup challenge response as JSON",
+                    )
+
+                return PasskeySignupChallengeResponse.model_validate(data)
+
+        except Exception as e:
+            if isinstance(e, (ApiError, MissingRequiredArgumentError, ValidationError)):
+                raise
+            raise ApiError("passkey_challenge_error", "Passkey signup challenge failed", e)
```
kishore7snehil: When we catch an unexpected error here (like a network timeout) and re-raise it as an ApiError, the original error gets hidden, the caller only sees "Passkey signup challenge failed" and can't tell it was actually a connection problem. Adding `from e (raise ApiError(...) from e)` keeps the original error linked in the traceback so it's still debuggable.  

Same applies to the equivalent line in `passkey_login_challenge` and `signin_with_passkey`.

kishore7snehil (changes requested):

src/auth0_server_python/auth_types/__init__.py
rmad17: 1. `__repr__` redaction  - I have not made any changes to pre existing patterns to avoid accidentaly breaking code in a large PR. Though I feel the redaction should be there in all cases. It does not cost us anything rather its a better approach in terms of security.  
2. `@model_validator (_check_at_least_one_method)` This was added for better clarity on the error rather than a generic 400 for better DX. However, this breaks push notifications - Removed it. 
3. `model_config = ConfigDict(populate_by_name=True)` - Copy paste mistake. Since it does not break anything this did not get caught in tests. 
4. `EnrollAuthenticationMethodRequest.type` - Added literals.
5. Updated the underscore to hyphens .

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2582,3 +2584,350 @@ async def login_with_custom_token_exchange(
...
+                    if not origin_issuer:
+                        raise IssuerValidationError(
+                            "Issuer missing from OIDC metadata. Cannot validate ID token issuer."
+                        )
+                    token_issuer = claims.get("iss", "")
+                    if self._normalize_url(token_issuer) != self._normalize_url(origin_issuer):
+                        raise IssuerValidationError(
+                            "ID token issuer mismatch. Ensure your Auth0 domain is configured correctly."
+                        )
+                    user_claims = UserClaims.model_validate(claims)
```
nandan-bhat, line 2978: We forward `organization` to the passkey grant above, but here we decode the ID token and never validate the org claim. The `complete_interactive_login` path in this same PR calls `validate_org_claims(claims, expected_org)` for exactly this reason, and the README says org claim validation is enforced automatically at callback.

As it stands, a passkey login pinned to an `organization` does not actually verify the returned token belongs to that org. In a multi org tenant that is an authorization gap. Can we add a `validate_org_claims` call here when `organization` was passed, and a test that asserts a mismatch raises `OrganizationTokenValidationError` and no session is stored ?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2582,3 +2584,350 @@ async def login_with_custom_token_exchange(
...
+                "client_id": self._client_id,
+                "auth_session": auth_session,
+                "authn_response": authn_response.model_dump(by_alias=True, exclude_none=True),
+            }
+            if self._client_secret:
+                body["client_secret"] = self._client_secret
+            if connection:
+                body["realm"] = connection
+            if organization:
+                body["organization"] = organization
```
nandan-bhat: The constructor stores `self._organization` and `start_interactive_login` uses `options.organization or self._organization`. The passkey methods only read the local `organization` argument, so the client level default is silently ignored here.

This means a client configured with a default org gets it applied to interactive logins but not to passkey logins, which is surprising. Can we apply the `self._organization` fallback here too so the behavior is consistent ?

src/auth0_server_python/auth_schemes/dpop_auth.py
```diff
@@ -0,0 +1,94 @@
...
+        response = yield request
+
+        # RFC 9449 §8.2 — server-nonce retry
+        if (
+            response.status_code == 401
+            and response.headers.get("DPoP-Nonce")
+        ):
+            nonce = response.headers["DPoP-Nonce"]
+            request.headers["DPoP"] = self._make_proof(request.method, str(request.url), nonce=nonce)
+            yield request
```
nandan-bhat, line 90: The nonce retry re-yields the same `request` object after it was already sent once. For the MyAccount calls that POST or PATCH a JSON body, I want to make sure the body is actually re-sent on the retry and not consumed.

The token endpoint path in `signin_with_passkey` sidesteps this by rebuilding the request with a fresh `client.post(...)`, but here we reuse the instance. On tenants that require a nonce the first call always 401s and retries, so this path runs every session. Could we add a test that drives a real bodied POST through `DPoPAuth` against a mock that returns 401 then 200, and asserts the second request still carries the original body ? If httpx does not re-stream it, we should rebuild the request instead.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2582,3 +2584,350 @@ async def login_with_custom_token_exchange(
...
+                    raise ApiError("token_expired", f"ID token has expired: {str(e)}", e)
+                except jwt.InvalidTokenError as e:
+                    raise ApiError("invalid_token", f"ID token verification failed: {str(e)}", e)
+
+            # Build token set and session state
+            token_set = TokenSet(
+                audience=audience or self.DEFAULT_AUDIENCE_STATE_KEY,
+                access_token=token_response.access_token,
+                scope=token_response.scope or scope or "",
+                expires_at=token_response.expires_at if token_response.expires_at is not None else int(time.time()) + token_response.expires_in,
```
nandan-bhat: Small one. `expires_in` is a required `int` on `PasskeyTokenResponse`, so by the time we get here `token_response.expires_in` is always an int and the earlier block already backfilled `expires_at`. The `else int(time.time()) + token_response.expires_in` branch cannot really be reached with a `None`, so this reads as dead defensive code.

If a server that returns only `expires_at` without `expires_in` is a real case, the model would reject it before this line anyway. Can we either make `expires_in` optional and keep one fallback, or drop the redundant branch ?

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -262,6 +287,7 @@ class CustomTokenExchangeOptions(BaseModel):
         organization: Organization identifier for the token exchange (optional)
         authorization_params: Additional OAuth parameters (optional)
     """
+
     subject_token: str
```
nandan-bhat, line 337: This drops the `field_validator` and `model_validator` (Bearer prefix, whitespace, `actor_token_type`) and the `min_length=1` constraint, and the checks were moved inline into `custom_token_exchange`. That means anyone building this model directly no longer gets validation at construction, and it only fires inside that one method.

Can we confirm `login_with_custom_token_exchange` still enforces the same rules on its own path ? If there was a specific reason for moving these out of the model (for example a Pydantic v2 change), it would help to note it in the PR description. Otherwise keeping them on the model is entry point agnostic and avoids duplicating the logic.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2582,3 +2584,350 @@ async def login_with_custom_token_exchange(
...
+            username: Optional username hint for conditional UI.
+            connection: Auth0 database connection name (realm).
+            organization: Auth0 organization ID or name.
+            store_options: Optional options for domain resolution.
+
+        Returns:
+            PasskeyLoginChallengeResponse with auth_session and authn_params_public_key.
+
+        Raises:
+            ApiError: If the challenge request fails.
```
nandan-bhat: The `Raises` section says `ApiError`, but the method body raises `PasskeyError` on a failed challenge. `passkey_signup_challenge` documents `PasskeyError` correctly. Can we fix this to `PasskeyError` ?

src/auth0_server_python/auth_types/__init__.py
rmad17: Will raise a seperate PR for CTE related fix.

nandan-bhat (dismissed):

README.md
```diff
@@ -177,7 +177,78 @@ The SDK handles per-domain OIDC discovery, JWKS fetching, issuer validation, and
...
+# Start enrolling a passkey (then sign it in the browser and verify)
+challenge = await my_account.enroll_authentication_method(
+    access_token=access_token,
+    request=EnrollAuthenticationMethodRequest(type="passkey"),
+)
+```
+
+For the full enroll/verify ceremony, listing, updating, deleting, and error handling, see [examples/MyAccountAuthenticationMethods.md](examples/MyAccountAuthenticationMethods.md).
+
+### 8. DPoP — Sender-Constrained Tokens
```
kishore7snehil: We should explicitly mention that this is with passkeys in the heading so that the users aren't confused or exclusively keep this section in `Passkeys.md`?

examples/Passkeys.md
```diff
@@ -0,0 +1,211 @@
...
+
+## How the flow works
+
+A passkey ceremony is always **two steps**, because the WebAuthn signature happens in the browser between them:
+
+1. **Challenge** — the SDK asks Auth0 for a challenge (`passkey_signup_challenge` / `passkey_login_challenge`). Auth0 returns an `auth_session` and the WebAuthn options (`authn_params_public_key`).
+2. **Browser** — your front end passes those options to `navigator.credentials.create()` (signup) or `navigator.credentials.get()` (login). The authenticator produces a signed credential.
+3. **Verify / sign-in** — the SDK exchanges the signed credential for tokens (`signin_with_passkey`) and **creates a server-side session**, exactly like every other login path.
+
+```
```
kishore7snehil: This diagram can be dropped.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2627,3 +2629,352 @@ async def login_with_custom_token_exchange(
...
+    ) -> PasskeyLoginResult:
+        """
+        Completes passkey authentication by exchanging the WebAuthn assertion
+        for tokens and establishing a server-side session.
+
+        This is step 2 of 2: call passkey_signup_challenge or passkey_login_challenge
+        first to obtain auth_session and the WebAuthn challenge options.
+
+        Uses Content-Type: application/json (required for nested authn_response).
+        Persists the session to the state store (same as complete_interactive_login).
```
kishore7snehil: We can drop this section of the docstring to make it consistent with other functions

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2627,3 +2629,352 @@ async def login_with_custom_token_exchange(
...
+                    if not origin_issuer:
+                        raise IssuerValidationError(
+                            "Issuer missing from OIDC metadata. Cannot validate ID token issuer."
+                        )
+                    token_issuer = claims.get("iss", "")
+                    if self._normalize_url(token_issuer) != self._normalize_url(origin_issuer):
+                        raise IssuerValidationError(
+                            "ID token issuer mismatch. Ensure your Auth0 domain is configured correctly."
+                        )
+                    if resolved_org:
```
kishore7snehil, line 2976: Here org membership is only checked inside the `if token_response.id_token:` block. `complete_interactive_login` also guards the case where an org was requested but the response carries no id_token (and no userinfo), it raises `OrganizationTokenValidationError` rather than establishing the session. On this path, if a caller passes `organization` but Auth0 returns only an access token with no id_token, we persist a session without ever verifying org membership.

Should we mirror the interactive-login guard here so both paths fail closed the same way?

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -5158,37 +5202,1678 @@ async def _fake_fetch(self, domain):
...
+    mocker.patch.object(client, "_verify_and_decode_jwt", return_value={
+        "sub": "auth0|user123", "iss": "https://auth0.local/"
+    })
+    mock_post = mocker.patch("httpx.AsyncClient.post", new_callable=AsyncMock)
+    mock_response = AsyncMock()
+    mock_response.status_code = 200
+    mock_response.json = MagicMock(return_value=_PASSKEY_TOKEN_RESPONSE)
+    mock_post.return_value = mock_response
+
+    with pytest.raises(Exception) as exc:
```
kishore7snehil: This catches broad `Exception` and only checks that "issuer" shows up in the message, though the docstring says it must raise `IssuerValidationError`. It would pass on an unrelated `KeyError`/`AttributeError` whose message happens to mention issuer. `pytest.raises(IssuerValidationError)` would pin the actual contract.

A couple of the passkey error tests nearby also lean on `assert X in str(exc) or Y in str(exc)`, asserting on `exc.value.code` instead would keep them from passing under the wrong failure mode.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -551,24 +630,224 @@ class MfaVerifyResponse(BaseModel):
...
+    refresh_token: Optional[str] = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def _backfill_expires_at(cls, data: Any) -> Any:
+        if isinstance(data, dict) and "expires_at" not in data and "expires_in" in data:
+            data["expires_at"] = int(time.time()) + int(data["expires_in"])
+        return data
+
+    def __repr__(self) -> str:
```
kishore7snehil: Redaction was added here on the new models, but `TokenSet` and `MfaVerifyResponse` in this same file hold `access_token` / `refresh_token` / `id_token` with no `__repr__` redaction. Having it on only the new models leaves the surface half-covered — worth deciding this one way project-wide.

src/auth0_server_python/auth_schemes/dpop_auth.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _make_proof(self, method: str, url: str, nonce: str = None) -> str:
```
kishore7snehil: `nonce: str = None` — the annotation says `str` but the default is `None`; `Optional[str]` matches the style used elsewhere in this file. Same on `make_dpop_proof_for_token_endpoint` above.

examples/MyAccountAuthenticationMethods.md
```diff
@@ -0,0 +1,237 @@
+# My Account API — Authentication Methods & Factors
```
kishore7snehil, line 1: Are we covering the already existing endpoints in this document as well? If not, let's add them with proper links to Auth0 Docs so that this file has a complete posture.

src/auth0_server_python/auth_types/__init__.py
kishore7snehil: > Can we confirm login_with_custom_token_exchange still enforces the same rules on its own path ? 

@nandan-bhat  Yes, it does.

kishore7snehil (changes requested):

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2448,7 +2447,10 @@ async def custom_token_exchange(
                         # before trusting any claim from the token.
                         if self._normalize_url(claims.get("iss", "")) == self._normalize_url(metadata.get("issuer")):
                             token_response.act = claims.get("act")
-                    except Exception:
+                    except (jwt.InvalidTokenError, ValueError, KeyError):
+                        # A genuinely absent/optional act claim or a benign decode
+                        # gap leaves act None. Anything outside these types (an
+                        # unexpected verify failure) surfaces rather than being masked.
```
kishore7snehil: We can trim down this comment to a crisp version.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -1188,18 +1187,18 @@ async def get_token_by_refresh_token(self, options: dict[str, Any]) -> dict[str,
...
+                        # no mfa_token is present (then falls through to ApiError).
+                        # store_pending is left False here: the get_access_token
+                        # caller persists the token in its own catch block.
+                        await self._mfa_client._raise_mfa_required(
+                            error_data,
+                            audience=audience or self.DEFAULT_AUDIENCE_STATE_KEY,
+                            scope=merged_scope or "",
+                            default_description="MFA required",
                         )
-                        error.mfa_token = error_data.get("mfa_token")
```
kishore7snehil, line 1197: This seems to be like a major strip down. What is the intention or alternate flow to achieve this?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -1100,24 +1114,9 @@ async def get_access_token(
...
-                        audience=audience or self.DEFAULT_AUDIENCE_STATE_KEY,
-                        scope=merged_scope or "",
-                        mfa_requirements=mfa_requirements
-                    )
-                    raise MfaRequiredError(
-                        "Multifactor authentication required",
-                        mfa_token=encrypted_token,
-                        mfa_requirements=mfa_requirements
-                    )
-
```
kishore7snehil, line 1120: This seems to be like a major strip down. What is the intention or alternate flow to achieve this?

kishore7snehil (changes requested):

README.md
```diff
@@ -177,7 +177,81 @@ The SDK handles per-domain OIDC discovery, JWKS fetching, issuer validation, and
...
+)
+
+my_account = MyAccountClient(domain=YOUR_CUSTOM_DOMAIN)
+
+# Start enrolling a passkey (then sign it in the browser and verify)
+challenge = await my_account.enroll_authentication_method(
+    access_token=access_token,
+    request=EnrollAuthenticationMethodRequest(type="passkey"),
+)
+```
```
kishore7snehil: Is the same code snippet which is already there in the .md file. If yes, please omit it from here as it will be a dupllication

README.md
```diff
@@ -177,7 +177,81 @@ The SDK handles per-domain OIDC discovery, JWKS fetching, issuer validation, and
...
+
+# Step 3 — complete sign-in and establish the session
+result = await auth0.signin_with_passkey(
+    auth_session=challenge.auth_session,
+    authn_response=PasskeyAuthResponse(**credential),
+    store_options={"request": request, "response": response}
+)
+
+user = result.state_data["user"]
+```
```
kishore7snehil: Same as [here](https://github.com/auth0/auth0-server-python/pull/116/changes#r3530425150)

kishore7snehil (changes requested):

README.md
```diff
@@ -177,7 +177,81 @@ The SDK handles per-domain OIDC discovery, JWKS fetching, issuer validation, and
 
 For more details and examples, see [examples/MultipleCustomDomains.md](examples/MultipleCustomDomains.md).
 
-### 6. Session Expiry from the Upstream IdP
```
kishore7snehil, line 180: Usually, we follow the sequence of releases to append the features to Readme. It's fine for now (beacuse session expiry is not a big feature as compared to Passkeys) but it's chronological and sync with the releases.

kishore7snehil (changes requested):

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -551,24 +630,224 @@ class MfaVerifyResponse(BaseModel):
...
+        None, alias="pubKeyCredParams"
+    )
+    authenticator_selection: Optional[PasskeyAuthenticatorSelection] = Field(
+        None, alias="authenticatorSelection"
+    )
+    timeout: Optional[int] = None
+    user_verification: Optional[str] = Field(None, alias="userVerification")
+
+
+EnrollmentType = Literal["passkey", "email", "phone", "totp", "push-notification", "recovery-code", "password"]
```
kishore7snehil: Why are we creating these variables here? It's used only once and can easily be integrated to the parameter?

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -551,24 +630,224 @@ class MfaVerifyResponse(BaseModel):
...
+    family_name: Optional[str] = None
+    nickname: Optional[str] = None
+    picture: Optional[str] = None
+
+
+class _PasskeyChallengeResponseBase(BaseModel):
+    auth_session: str
+    authn_params_public_key: PasskeyPublicKeyOptions
+
+    def __repr__(self) -> str:
```
kishore7snehil: Redaction code added which needs to be removed

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -551,24 +630,224 @@ class MfaVerifyResponse(BaseModel):
...
+    connection: Optional[str] = None
+
+
+class EnrollmentChallengeResponse(BaseModel):
+    model_config = ConfigDict(extra="allow")
+    authentication_method_id: str
+    auth_session: str
+    authn_params_public_key: Optional[PasskeyPublicKeyOptions] = None
+
+    def __repr__(self) -> str:
```
kishore7snehil: Another redaction code added, which need to be removed

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -551,24 +630,224 @@ class MfaVerifyResponse(BaseModel):
...
+    model_config = ConfigDict(extra="allow")
+    access_token: str
+    token_type: str = "Bearer"
+    expires_in: int
+    expires_at: int
+    scope: Optional[str] = None
+    id_token: Optional[str] = None
+    refresh_token: Optional[str] = None
+
+    @model_validator(mode="before")
```
kishore7snehil: Every other token path in the SDK computes `expires_at` in the client method right after the HTTP response (see the `int(time.time()) + expires_in` spots in `server_client.py`). This is the only place that backfills it inside the model via a `mode="before"` validator. We should be consistent.

kishore7snehil (changes requested):

README.md
kishore7snehil: @rmad17 Is this intentional?

kishore7snehil (changes requested):

src/auth0_server_python/auth_schemes/dpop_auth.py
```diff
@@ -0,0 +1,99 @@
...
+            raise ValueError("Access token must contain only ASCII characters")
+        self._token = token
+        self._key = key
+        self._public_jwk = public_jwk
+
+    def __repr__(self) -> str:
+        return "DPoPAuth(token=[REDACTED], key=[REDACTED])"
+
+    def __str__(self) -> str:
+        return "DPoPAuth(token=[REDACTED], key=[REDACTED])"
```
kishore7snehil: We should drop this.

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -130,13 +132,127 @@ def decrypt_mfa_token(self, encrypted_token: str) -> MfaTokenContext:
...
         if elapsed > DEFAULT_MFA_TOKEN_TTL:
             raise MfaTokenExpiredError()
 
         return context
 
+    # ============================================================================
+    # MFA STATE
+    # ============================================================================
+
+    async def store_pending_mfa(
```
kishore7snehil: I can see this function used only once. Can we not include the function's logic there itself? Do we anticipate need of this helper function in future?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -130,13 +132,127 @@ def decrypt_mfa_token(self, encrypted_token: str) -> MfaTokenContext:
...
+    ) -> None:
+        """Save an in-progress MFA token so challenge and verify can proceed without the client carrying the token."""
+        if self._state_store:
+            await self._state_store.set(
+                MFA_PENDING_IDENTIFIER,
+                {"mfa_token": encrypted_token},
+                options=store_options,
+            )
+
+    async def get_pending_mfa(
```
kishore7snehil: I don't see this function called anywhere.Am I missing something?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -130,13 +132,127 @@ def decrypt_mfa_token(self, encrypted_token: str) -> MfaTokenContext:
...
+    ) -> Optional[str]:
+        """Retrieve the in-progress MFA token if a challenge is pending for the current session, or None."""
+        if not self._state_store:
+            return None
+        data = await self._state_store.get(MFA_PENDING_IDENTIFIER, store_options)
+        if data and isinstance(data, dict):
+            return data.get("mfa_token")
+        return None
+
+    async def _clear_pending_mfa(
```
kishore7snehil: I can see this function used only once. Can we not include the function's logic there itself? Do we anticipate need of this helper function in future?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -130,13 +132,127 @@ def decrypt_mfa_token(self, encrypted_token: str) -> MfaTokenContext:
...
+        Encrypt the server-issued mfa_token and raise MfaRequiredError.
+
+        Shared by every site that handles an `mfa_required` response so the
+        encrypt-then-raise behaviour cannot drift between entry points. Returns
+        only when the response carries no mfa_token (caller then falls through
+        to its own typed error).
+
+        store_pending controls whether the encrypted token is persisted to the
+        state store before raising. It is an explicit argument so the difference
+        between entry points is visible: the passkey grant persists it here,
```
kishore7snehil: The docstring is verbose and doesn't follow the standard format of 

- Crisp Description of the fucntion (what it does)
- Args
- Returns/Raises

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -130,13 +132,127 @@ def decrypt_mfa_token(self, encrypted_token: str) -> MfaTokenContext:
...
+
+    @staticmethod
+    def _parse_error_body(response: httpx.Response) -> dict[str, Any]:
+        """
+        Parse an error response body as JSON, falling back to a status-coded
+        stub when the body is not JSON (e.g. a gateway 502/504 HTML page).
+
+        Never raises — the caller always gets a dict it can read error fields
+        from, so a non-JSON error surfaces the real HTTP status rather than a
+        JSON-parser exception folded into the message.
```
kishore7snehil: The docstring doesn't follow the standard format of

- Crisp Description of the fucntion (what it does)
- Args
- Returns/Raises

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -159,19 +275,19 @@ async def list_authenticators(
         Raises:
             MfaListAuthenticatorsError: When the request fails.
         """
-        mfa_token = options["mfa_token"]
+        context = self.decrypt_mfa_token(self._resolve_encrypted_token(options))
```
kishore7snehil: Wy have we added a function(`_resolve_encrypted_token`) for just one validation check i.e. if token is present or not? 

This should have been handled here itself.

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -207,7 +323,7 @@ async def enroll_authenticator(
         Raises:
             MfaEnrollmentError: When enrollment fails.
         """
-        mfa_token = options["mfa_token"]
+        context = self.decrypt_mfa_token(self._resolve_encrypted_token(options))
```
kishore7snehil: Same as [here](https://github.com/auth0/auth0-server-python/pull/116/changes#r3536128852)

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -292,7 +408,7 @@ async def challenge_authenticator(
         Raises:
             MfaChallengeError: When the challenge fails.
         """
-        mfa_token = options["mfa_token"]
+        context = self.decrypt_mfa_token(self._resolve_encrypted_token(options))
```
kishore7snehil: Same as [here](https://github.com/auth0/auth0-server-python/pull/116/changes#r3536128852)

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -373,13 +489,12 @@ async def verify(
             MfaVerifyError: When verification fails.
             MfaRequiredError: When chained MFA is required.
         """
-        mfa_token = options["mfa_token"]
+        context = self.decrypt_mfa_token(self._resolve_encrypted_token(options))
```
kishore7snehil: Same as [here](https://github.com/auth0/auth0-server-python/pull/116/changes#r3536128852)

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -270,8 +386,8 @@ async def enroll_authenticator(
             raise
         except Exception as e:
             raise MfaEnrollmentError(
-                f"Unexpected error enrolling authenticator: {str(e)}"
-            )
+                "Unexpected error enrolling authenticator"
```
kishore7snehil, line 375: Are we loosing the exact error message store in `e` here? Please check once

src/auth0_server_python/auth_schemes/dpop_auth.py
```diff
@@ -0,0 +1,99 @@
...
+
+def _validate_dpop_key(key: "jwk.JWK") -> dict:
+    """Return the public JWK after enforcing the EC P-256 requirement (ES256)."""
+    public_jwk = key.export_public(as_dict=True)
+    if public_jwk.get("kty") != "EC" or public_jwk.get("crv") != "P-256":
+        raise ValueError("DPoP key must be an EC P-256 key")
+    return public_jwk
+
+
+def make_dpop_proof_for_token_endpoint(
```
kishore7snehil: This is nearly a line-for-line duplicate of [DPoPAuth._make_proof](https://github.com/auth0/auth0-server-python/pull/116/changes#diff-30fa236c7406cddcbca03fa6d046013ad872eee4b91a4700409ce913d05543beR82) below

kishore7snehil (changes requested):

examples/MyAccountAuthenticationMethods.md
rmad17: Did not find any new endpoints to add.

src/auth0_server_python/auth_types/__init__.py
rmad17: This is a cleaner approach. Any reuse of this in future will not require rework. 
Will move it to the top of the file.

src/auth0_server_python/auth_types/__init__.py
rmad17: Redaction is a good security practice and we should use it. Do you see any problem with redaction? 
The issue I see here is `str(enrollmentChallengeResponse())` would still leak the secret as it will fallback to __str__ of `BaseModel`. For that I will add `Field(repr=False)` in the attribute of the Schema.

src/auth0_server_python/auth_types/__init__.py
rmad17: Will perform the redaction via `Field` param.

src/auth0_server_python/auth_types/__init__.py
rmad17: We cannot achieve it project wide in one go. Isn;t it a good idea to start and improve the other models too as we go forward. Its not like these are interdependent.

src/auth0_server_python/auth_server/mfa_client.py
rmad17: This would mean writing the following 4 times
```python
token = options.get("mfa_token")
  if not token:
      raise MfaTokenInvalidError()
```

kishore7snehil: Yes and that's fine. It would have made more sense if the validation cases list was longer

kishore7snehil (changes requested):

examples/MyAccountAuthenticationMethods.md
```diff
@@ -0,0 +1,241 @@
...
+> Method IDs (e.g. `passkey|abc123`) can contain characters like `|`. The SDK URL-encodes every ID it places in a path, so pass the raw ID exactly as returned — do not pre-encode it.
+
+## 5. Update (rename) an authentication method
+
+```python
+from auth0_server_python.auth_types import UpdateAuthenticationMethodRequest
+
+method = await my_account.update_authentication_method(
+    access_token=access_token,
+    authentication_method_id="passkey|abc123",
```
pmathew92: name updation on passkey is not supported . Lets not add an wrong id example here. Add something like `totp|123`

examples/Passkeys.md
```diff
@@ -0,0 +1,202 @@
...
+    store_options={"request": request, "response": response},
+)
+
+# Hand these to the browser:
+#   challenge.auth_session              -> opaque session credential (Tier 1, do not log)
+#   challenge.authn_params_public_key   -> pass to navigator.credentials.create()
+```
+
+> [!TIP]
+> `PasskeyUserProfile` allows extra fields — any additional profile attribute your tenant accepts (for example `given_name`, `family_name`, `picture`) passes through without an SDK change. Pass tenant-specific custom data via the separate `user_metadata` argument.
```
pmathew92, line 72: Better showcase an example with all the supported properties ?

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -336,13 +392,28 @@ class LoginWithCustomTokenExchangeResult(BaseModel):
...
+    Result from signin_with_passkey.
+
+    Contains the session data established after the webauthn token exchange.
+    Mirrors LoginWithCustomTokenExchangeResult — passkey sign-in is a complete
+    login ceremony and creates a server-side session like every other login path.
+    """
+
+    state_data: dict[str, Any]
+
+
```
kishore7snehil: Please move this to the Passkey section below

src/auth0_server_python/error/__init__.py
```diff
@@ -256,6 +256,24 @@ class CustomTokenExchangeErrorCode:
...
+        self.cause = cause
+
+
+class PasskeyErrorCode:
+    """Error codes for passkey operations."""
+    CHALLENGE_FAILED = "passkey_challenge_error"
+    TOKEN_EXCHANGE_FAILED = "passkey_token_error"
+    INVALID_RESPONSE = "invalid_response"
+
+
```
kishore7snehil: This should be at the bottom and should have a section of it's own.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -5186,9 +5231,1612 @@ async def _fake_fetch(self, domain):
 
 
 # =============================================================================
-# Organization and Invitation Tests
+# PASSKEY AUTHENTICATION
```
kishore7snehil: Again the placement of this section is incorrect, please move it after Organization and Invitation tests.

kishore7snehil (changes requested):

src/auth0_server_python/auth_schemes/dpop_auth.py
```diff
@@ -0,0 +1,94 @@
...
+        request.headers["Authorization"] = f"DPoP {self._token}"
+        request.headers["DPoP"] = proof
+        response = yield request
+
+        # RFC 9449 §8.2 — server-nonce retry
+        if response.status_code == 401 and response.headers.get("DPoP-Nonce"):
+            nonce = response.headers["DPoP-Nonce"]
+            request.headers["DPoP"] = self._make_proof(
+                request.method, str(request.url), nonce=nonce
+            )
```
pmathew92, line 89: Should we set up  retry limit cap here?

rmad17: The auth_flow generator yields at most twice.

src/auth0_server_python/auth_server/mfa_client.py
rmad17: `raise MfaEnrollmentError(...)` from `e` preserves the full exception via the __cause__ chain.

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -365,21 +471,31 @@ async def verify(
                 - 'audience': str (optional, required if persist=True) - Audience for token_set
                 - 'scope': str (optional) - Scope for token_set
             store_options: Optional options passed to the State Store (e.g. request/response).
+            dpop_key: Optional EC P-256 JWK for DPoP-bound token exchange. Pass the
```
kishore7snehil: It reads too verbose - we can trim this down

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -403,28 +519,46 @@ async def verify(
...
+                        dpop_key, "POST", token_endpoint
+                    )
                 response = await client.post(
                     token_endpoint,
                     data=body,
-                    headers={"Content-Type": "application/x-www-form-urlencoded"}
+                    headers=headers
                 )
 
+                # RFC 9449 §8.2 — the authorization server signals a required
```
kishore7snehil: Verbose inline comment

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -435,11 +569,33 @@ async def verify(
...
+                    # possession on later calls; reject rather than downgrade.
+                    raise MfaVerifyError(
+                        "DPoP token binding failed: expected token_type 'DPoP', "
+                        f"got '{verify_response.token_type}'"
+                    )
+                if dpop_key is None and token_is_dpop:
+                    # Server issued a DPoP-bound token but no key was supplied —
+                    # we cannot prove possession, so accepting it would fail open.
+                    raise MfaVerifyError(
+                        "Server returned a DPoP-bound token but no dpop_key was "
```
kishore7snehil: Is this claim ("Server returned a DPoP-bound toke")verfied? Did you check the Auth0 Server response in this case?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -435,11 +569,33 @@ async def verify(
...
+                        "provided; pass dpop_key to verify to bind it"
```
kishore7snehil: We can modify the wording a bit

kishore7snehil (changes requested):

src/auth0_server_python/tests/test_mfa_client.py
```diff
@@ -843,6 +890,124 @@ async def test_verify_persist_store_failure_raises(self, mocker):
...
+        dpop_key = jwk.JWK.generate(kty="EC", crv="P-256")
+        response = AsyncMock()
+        response.status_code = 200
+        response.headers = {}
+        response.json = MagicMock(return_value={
+            "access_token": "at", "token_type": "Bearer", "expires_in": 3600
+        })
+        mocker.patch("httpx.AsyncClient.post", new_callable=AsyncMock, return_value=response)
+
+        with pytest.raises(MfaVerifyError, match="DPoP token binding failed"):
```
kishore7snehil, line 971: What does the Auth0 Server respond with in this case?

kishore7snehil (changes requested):

nandan-bhat:
@rmad17
Ideally, it might be better to split this into three separate PRs, since `Passkeys`, `My Account`, and `DPoP` support are all major features and would be easier to review, test, and track independently.

Given the number of review comments and discussions already captured in this PR, perhaps we can first address the remaining feedback here and then split the implementation into three separate PRs. Each new PR can reference this one and note that the implementation has already been reviewed here.

src/auth0_server_python/auth_server/mfa_client.py
rmad17: On validation found its not what Auth0 server says. Its unreachable code. Removing this.

src/auth0_server_python/auth_server/mfa_client.py
rmad17: Removed it as the section was dead code.

src/auth0_server_python/tests/test_mfa_client.py
rmad17: HTTP 200 + token_type: "Bearer". 
This is more of a defensive gaurdrail test to ensure all required steps related to token constraint are met and ensures the SDK never silently downgrades the token from DPoP to Bearer.

PR #117, opened 2026-06-04
PoC: enforce upstream IdP session_expiry ceiling

kishore7snehil:
Adds enforcement of the IPSIE `session_expiry` claim for enterprise connections.
When the connection is configured to honor an upstream IdP session ceiling, Auth0
includes a `session_expiry` claim (absolute Unix seconds) in the ID token. The SDK
reads it at login, stores it with the session, and enforces it on every session read.

### Enforcement
- Reads `session_expiry` from userinfo / verified ID-token claims at
  `complete_interactive_login` and stamps it onto the internal session state.
- `get_user()` and `get_session()` return `None` once the ceiling is reached
  (silent — behaves like no session, so existing redirect-to-login fires).
- `get_access_token()` raises `AccessTokenError` with code `session_expired`
  (loud — checked before serving cache or refreshing; refresh is never attempted).
- Reaching the ceiling deletes the stored session before returning.
- 30s negative leeway for clock skew; integer-seconds comparison.
- Refresh-token grant preserves the original ceiling (not re-emitted on refresh).

### Claims access
- `session_expiry` is surfaced on `UserClaims`, so it can be read via
  `get_user()` without triggering enforcement.

### New / changed types & errors
- `UserClaims.session_expiry: Optional[int]`
- `InternalStateData.session_expires_at: Optional[int]`
- `AccessTokenErrorCode.SESSION_EXPIRED = "session_expired"`

### DIFF
--- a/src/auth0_server_python/auth_server/server_client.py
+++ b/src/auth0_server_python/auth_server/server_client.py
@@ -648,8 +648,8 @@ class ServerClient(Generic[TStoreOptions]):
             user_claims = UserClaims.parse_obj(user_info)
             # authlib populates `userinfo` from parsed ID token claims, so the
             # IPSIE session_expiry claim may surface here.
-            session_expires_at = State.extract_epoch_claim(user_info, "session_expiry")
-            issued_at = State.extract_epoch_claim(user_info, "iat")
+            session_expires_at = user_info.get("session_expiry")
+            issued_at = user_info.get("iat")
         elif id_token:
             # Fetch JWKS for signature verification
             jwks = await self._get_jwks_cached(origin_domain, metadata)
@@ -667,8 +667,8 @@ class ServerClient(Generic[TStoreOptions]):
 
                 user_claims = UserClaims.parse_obj(claims)
                 # IPSIE session_expiry ceiling from the verified ID token.
-                session_expires_at = State.extract_epoch_claim(claims, "session_expiry")
-                issued_at = State.extract_epoch_claim(claims, "iat")
+                session_expires_at = claims.get("session_expiry")
+                issued_at = claims.get("iat")
             except ValueError as e:
                 raise ApiError("jwks_key_not_found", str(e))
             except jwt.InvalidSignatureError as e:
--- a/src/auth0_server_python/auth_server/server_client.py
+++ b/src/auth0_server_python/auth_server/server_client.py
@@ -638,9 +638,14 @@ class ServerClient(Generic[TStoreOptions]):
         user_info = token_response.get("userinfo")
         user_claims = None
         id_token = token_response.get("id_token")
+        # IPSIE session_expiry ceiling, read from the verified ID token claims.
+        session_expires_at = None
 
         if user_info:
             user_claims = UserClaims.parse_obj(user_info)
+            # authlib populates `userinfo` from parsed ID token claims, so the
+            # IPSIE session_expiry claim may surface here.
+            session_expires_at = State.extract_session_expiry(user_info)
         elif id_token:
             # Fetch JWKS for signature verification
             jwks = await self._get_jwks_cached(origin_domain, metadata)
@@ -657,6 +662,8 @@ class ServerClient(Generic[TStoreOptions]):
                     raise IssuerValidationError("ID token issuer mismatch. Ensure your Auth0 domain is configured correctly.")
 
                 user_claims = UserClaims.parse_obj(claims)
+                # IPSIE session_expiry ceiling from the verified ID token.
+                session_expires_at = State.extract_session_expiry(claims)
             except ValueError as e:
                 raise ApiError("jwks_key_not_found", str(e))
             except jwt.InvalidSignatureError as e:
@@ -708,7 +715,8 @@ class ServerClient(Generic[TStoreOptions]):
             domain=origin_domain,
             internal={
                 "sid": sid,
-                "created_at": int(time.time())
+                "created_at": int(time.time()),
+                "session_expires_at": session_expires_at
             }
         )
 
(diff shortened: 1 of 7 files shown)


src/auth0_server_python/auth_server/server_client.py
```diff
@@ -638,9 +639,17 @@ async def complete_interactive_login(
...
+        # IPSIE session_expiry ceiling, read from the verified ID token claims.
+        session_expires_at = None
+        # ID token `iat`, used to detect a ceiling that is already past at login.
+        issued_at = None
 
         if user_info:
             user_claims = UserClaims.parse_obj(user_info)
+            # authlib populates `userinfo` from parsed ID token claims, so the
+            # IPSIE session_expiry claim may surface here.
+            session_expires_at = user_info.get("session_expiry")
```
nandan-bhat, line 651: I want to make sure we are reading `session_expiry` from a trusted source here. In the `id_token` branch below we verify the signature and run the normalized issuer check before we trust any claim. On this `userinfo` branch we read `session_expiry` and `iat` straight off the dict without either of those checks. Since session_expiry is the thing that caps the session, trusting it from an unverified source feels risky. The comment says `authlib` fills `userinfo` from the parsed ID token, so does `authlib` actually verify that token's signature for us in every flow we support ? If we are not sure, I would rather pull `session_expiry` from the verified claims path so both branches are equally safe.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -685,6 +697,10 @@ async def complete_interactive_login(
                 )
 
 
+        # Refuse to persist a session whose ceiling is already in the past.
+        if State.is_session_ceiling_in_past(session_expires_at, issued_at):
+            raise SessionExpiredError()
```
nandan-bhat, line 702: When we raise here we skip the transaction cleanup that happens later in the method, so the transaction record is left behind. The auth code has already been spent at `fetch_token`, so the user cannot reuse this transaction anyway, a retry will start a brand new one. So the old record just sits there until it expires. Is leaving it intentional, or should we delete the transaction before raising so we do not pile up dead records ? The test comment says we leave it so the user can retry, but a retry does not reuse it, so that reasoning does not quite hold.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -972,6 +1014,15 @@ async def get_access_token(
 
         merged_scope = self._merge_scope_with_defaults(scope, audience)
 
+        # Once the session ceiling has passed, fail instead of serving or refreshing a token.
+        internal = (state_data_dict or {}).get("internal") or {}
+        if State.is_session_ceiling_reached(internal.get("session_expires_at")):
```
nandan-bhat, line 1019: This is the same ceiling check plus session delete that `_is_session_expired_by_ceiling` already does. The only difference is that here we raise instead of returning `None`. Could we have the helper do the check and delete, and just decide here whether to raise or return ? That way the leeway and delete behavior live in one place and cannot drift apart later.

src/auth0_server_python/utils/helpers.py
```diff
@@ -91,12 +95,20 @@ def update_state_data(
                     else ts
                     for ts in token_sets
                 ]
+            # Preserve the IPSIE session_expiry ceiling stamped at login. The
+            # platform does not re-emit session_expiry on a refresh-token grant
+            # (it doesn't round-trip the upstream IdP), so the value from the
+            # refreshed ID token must NOT overwrite or erase the original
+            # ceiling — doing so would let the session outlive its bound.
+            internal = dict(state_data_dict.get("internal") or {})
```
nandan-bhat, line 103: Just so I understand this line: the return below already spreads `state_data_dict`, which carries internal along with `session_expires_at`, and the refresh response has no path that writes into internal. So is this copy actually changing anything, or is it just a defensive copy ? The comment makes it sound like without this line the ceiling would get erased on refresh, but I do not think the old code ever touched internal. If it is defensive only, can we soften the comment so it does not imply a bug that was not there ?

src/auth0_server_python/utils/helpers.py
```diff
@@ -177,6 +189,35 @@ def update_state_data_for_connection_token_set(
...
+        True when the session ceiling is already in the past at login.
+
+        Compares the ceiling against the ID token `iat`, or wall-clock now when
+        `iat` is absent, using the same leeway as is_session_ceiling_reached. A
+        None ceiling means none was asserted and is never treated as expired.
+        """
+        if session_expires_at is None:
+            return False
+        reference = issued_at if issued_at else int(time.time())
+        return session_expires_at <= (reference + cls.SESSION_EXPIRY_LEEWAY_SECONDS)
```
nandan-bhat, line 218: Small thing: this uses `if issued_at`, so an iat of 0 would fall through to wall clock now. In practice iat is never 0, so this is not a real bug, but `if issued_at is not None` would read more clearly and match how we guard the ceiling itself.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -4816,3 +4817,439 @@ async def _fake_fetch(self, domain):
...
     finally:
         ServerClient._fetch_oidc_metadata = original_fetch
+
+
+# =============================================================================
+# IPSIE session_expiry enforcement
+# =============================================================================
+
+
+def test_is_session_ceiling_reached_none_never_expires():
```
nandan-bhat, line 4827: Nice coverage on the ceiling math and the read and login paths. Two gaps I noticed. First, nothing tests that a refresh keeps the original ceiling, which is the whole reason for the internal copy in helpers.py, so a regression there would slip through. Could we add a test that runs `update_state_data` with a refresh response and checks `session_expires_at` is still set ? Second, all the login tests go through the id_token branch. The userinfo branch that reads session_expiry is never exercised, and that is the branch I asked about on the security side. A test for it would help.

kishore7snehil:
Closing as superseded by #120

PR #120, opened 2026-06-11
feat: enforce upstream IdP session_expiry ceiling (IPSIE SL1)

kishore7snehil:
### 📋 Changes

This PR adds enforcement of the IPSIE SL1 `session_expiry` claim to auth0-server-python — an absolute Unix-epoch ceiling an upstream IdP can assert on how long an Auth0 app session may live. The SDK reads the claim at login, stamps it on `internal.session_expires_at`, and enforces it lazily on every read.

#### ✨ Features
- **Login-time enforcement:** `complete_interactive_login` reads `session_expiry`/`iat` from the verified ID token claims and stamps `internal.session_expires_at` on the persisted session. A login whose ceiling is already in the past is rejected with `SessionExpiredError` rather than persisted; the spent transaction is cleaned up before raising since its authorization code cannot be reused.
- **Read-time enforcement:** `get_session()` / `get_user()` return `None` (and delete the session) once the ceiling is reached; `get_access_token()` raises `AccessTokenError(SESSION_EXPIRED)`. A 30s negative clock-skew leeway is applied so the SDK never serves a session the platform has already revoked.
- **Refresh preservation:** The ceiling stamped at login is preserved across refresh-token grants — the platform does not re-emit `session_expiry` on a refresh, so the original ceiling is never overwritten or erased.

#### 🔧 API Changes
- New `SessionExpiredError` — flow-agnostic, carries `code == "session_expired"`
- `UserClaims.session_expiry: Optional[int]` and `InternalStateData.session_expires_at: Optional[int]`

#### 📖 Documentation
- Updated `README.md` and `examples/RetrievingData.md` with usage, the upgrade null-check note, leeway behavior, and refresh preservation

#### 🧪 Testing

### Contributor Checklist


nandan-bhat (approved):

nandan-bhat:
LGTM

PR #122, opened 2026-06-20
feat: validate CTE actor token pairing and surface act claim for delegation/impersonation

kishore7snehil:
### 📋 Changes

This PR completes Custom Token Exchange (CTE) Delegation/Impersonation support (RFC 8693) in auth0-server-python. CTE input validation moves into the exchange methods so callers get typed `CustomTokenExchangeError`s with specific codes, the `actor_token`/`actor_token_type` pairing is enforced in both directions, and the `act` claim from a delegation/impersonation exchange is surfaced to the developer.

#### ✨ Features
- **Actor-token pairing (symmetric):** `custom_token_exchange` and `login_with_custom_token_exchange` now reject `actor_token` without `actor_token_type` (`MISSING_ACTOR_TOKEN_TYPE`) and `actor_token_type` without `actor_token` (`MISSING_ACTOR_TOKEN`), matching how Auth0 enforces the pair.
- **`act` claim surfacing:** When an exchange includes an actor token, the SDK verifies the returned ID token and exposes the RFC 8693 §4.1 `act` claim on `TokenExchangeResponse.act` (stateless) and on the persisted session user, readable back via `get_user()` (session-creating).
- **Typed validation:** Empty/whitespace and `Bearer `-prefixed tokens now raise `CustomTokenExchangeError(INVALID_TOKEN_FORMAT)`.

#### 🔧 API Changes
- New error code `CustomTokenExchangeErrorCode.MISSING_ACTOR_TOKEN`
- `TokenExchangeResponse.act: Optional[dict[str, Any]]`
- CTE input validation now runs inside `custom_token_exchange` / `login_with_custom_token_exchange` and raises `CustomTokenExchangeError` with a specific code (`INVALID_TOKEN_FORMAT`, `MISSING_ACTOR_TOKEN_TYPE`, `MISSING_ACTOR_TOKEN`)

> ⚠️ **Minor breaking change — exception type on invalid input:** CTE input validation moved out of the Pydantic models and into the exchange methods. Previously, invalid input (e.g. an empty token) raised a Pydantic `ValidationError` at the point you constructed `CustomTokenExchangeOptions`. It now raises a `CustomTokenExchangeError` with a specific code at call time instead. If you were catching `ValidationError` for this, switch to catching `CustomTokenExchangeError`.

#### 📖 Documentation
- Updated `examples/CustomTokenExchange.md` with `act` read-back (stateless and session), the refresh-token suppression note, and the `MISSING_ACTOR_TOKEN` error code

#### 🧪 Testing

### Contributor Checklist

### DIFF
--- a/src/auth0_server_python/auth_server/server_client.py
+++ b/src/auth0_server_python/auth_server/server_client.py
@@ -2251,10 +2251,20 @@ class ServerClient(Generic[TStoreOptions]):
                     CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
                     "subject_token cannot be empty or whitespace-only"
                 )
+            if not options.subject_token_type.strip():
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "subject_token_type cannot be empty or whitespace-only"
+                )
             if options.subject_token.strip().startswith("Bearer "):
                 raise CustomTokenExchangeError(
                     CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
                     "subject_token should not include 'Bearer ' prefix"
+                )
+            if options.actor_token is not None and not options.actor_token.strip():
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "actor_token cannot be empty or whitespace-only"
                 )
             if options.actor_token and options.actor_token.strip().startswith("Bearer "):
                 raise CustomTokenExchangeError(
@@ -2444,6 +2454,8 @@ class ServerClient(Generic[TStoreOptions]):
                             "ID token issuer mismatch. Ensure your Auth0 domain is configured correctly."
                         )
 
+                    # UserClaims allows extra fields, so any act claim in the
+                    # verified id_token is carried onto the session user here.
                     user_claims = UserClaims.parse_obj(claims)
                     # Extract sid from token if available
                     sid = claims.get("sid", sid)
--- a/src/auth0_server_python/auth_server/server_client.py
+++ b/src/auth0_server_python/auth_server/server_client.py
@@ -2336,13 +2336,21 @@ class ServerClient(Generic[TStoreOptions]):
 
                 token_response = TokenExchangeResponse(**token_data)
 
-                # Surface the actor claim for delegation exchanges
+                # Surface the actor claim for delegation exchanges. Best-effort:
+                # a decode/verify hiccup must not fail an exchange the token
+                # endpoint already accepted, so act stays None on any failure.
                 if options.actor_token and token_response.id_token:
-                    jwks = await self._get_jwks_cached(domain, metadata)
-                    claims = await self._verify_and_decode_jwt(
-                        token_response.id_token, jwks, audience=self._client_id
-                    )
-                    token_response.act = claims.get("act")
+                    try:
+                        jwks = await self._get_jwks_cached(domain, metadata)
+                        claims = await self._verify_and_decode_jwt(
+                            token_response.id_token, jwks, audience=self._client_id
+                        )
+                        # Apply the same normalized issuer check the login path uses
+                        # before trusting any claim from the token.
+                        if self._normalize_url(claims.get("iss", "")) == self._normalize_url(metadata.get("issuer")):
+                            token_response.act = claims.get("act")
+                    except Exception:
+                        token_response.act = None
 
                 return token_response
 
--- a/src/auth0_server_python/auth_server/server_client.py
+++ b/src/auth0_server_python/auth_server/server_client.py
@@ -2243,9 +2243,34 @@ class ServerClient(Generic[TStoreOptions]):
             https://datatracker.ietf.org/doc/html/rfc8693
         """
         try:
-            # Validate options (Pydantic handles this automatically)
             if not isinstance(options, CustomTokenExchangeOptions):
                 options = CustomTokenExchangeOptions(**options)
+
+            if not options.subject_token.strip():
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "subject_token cannot be empty or whitespace-only"
+                )
+            if options.subject_token.strip().startswith("Bearer "):
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "subject_token should not include 'Bearer ' prefix"
+                )
+            if options.actor_token and options.actor_token.strip().startswith("Bearer "):
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "actor_token should not include 'Bearer ' prefix"
+                )
+            if options.actor_token and not options.actor_token_type:
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.MISSING_ACTOR_TOKEN_TYPE,
+                    "actor_token_type is required when actor_token is provided"
+                )
+            if options.actor_token_type and not options.actor_token:
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.MISSING_ACTOR_TOKEN,
+                    "actor_token is required when actor_token_type is provided"
+                )
 
             # Resolve domain
             domain = await self._resolve_current_domain(store_options)
@@ -2309,8 +2334,17 @@ class ServerClient(Generic[TStoreOptions]):
                         "Failed to parse token response as JSON"
                     )
 
-                # Validate and return response
-                return TokenExchangeResponse(**token_data)
+                token_response = TokenExchangeResponse(**token_data)
+
+                # Surface the actor claim for delegation exchanges
+                if options.actor_token and token_response.id_token:
+                    jwks = await self._get_jwks_cached(domain, metadata)
+                    claims = await self._verify_and_decode_jwt(
+                        token_response.id_token, jwks, audience=self._client_id
+                    )
+                    token_response.act = claims.get("act")
+
+                return token_response
 
         except ValidationError as e:
             raise CustomTokenExchangeError(
--- a/src/auth0_server_python/auth_types/__init__.py
+++ b/src/auth0_server_python/auth_types/__init__.py
@@ -5,7 +5,7 @@ These Pydantic models provide type safety and validation for all SDK data struct
 
 from typing import Any, Literal, Optional, Union
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import BaseModel, Field
 
 
 class UserClaims(BaseModel):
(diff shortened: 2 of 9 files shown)


src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2243,10 +2243,35 @@ async def custom_token_exchange(
             https://datatracker.ietf.org/doc/html/rfc8693
         """
         try:
-            # Validate options (Pydantic handles this automatically)
             if not isinstance(options, CustomTokenExchangeOptions):
                 options = CustomTokenExchangeOptions(**options)
 
+            if not options.subject_token.strip():
```
nandan-bhat, line 2293: Looks like we dropped the min_length check on subject_token_type when we removed the Pydantic validators, but we did not add a replacement check here. So an empty subject_token_type now passes and gets sent to the token endpoint, which comes back as a generic TOKEN_EXCHANGE_FAILED instead of a clean local error. We do check subject_token right above, so this feels a bit inconsistent. Can we add the same empty/whitespace check for subject_token_type ?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2243,10 +2243,35 @@ async def custom_token_exchange(
...
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "subject_token cannot be empty or whitespace-only"
+                )
+            if options.subject_token.strip().startswith("Bearer "):
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "subject_token should not include 'Bearer ' prefix"
+                )
+            if options.actor_token and options.actor_token.strip().startswith("Bearer "):
```
nandan-bhat, line 2313: The old validator rejected whitespace-only actor_token as well, but here we only check it for the Bearer prefix. A value like "   " is still truthy, so it slips past the pairing check and gets sent to Auth0 as is. Can we add an empty/whitespace check for actor_token so it matches what we do for subject_token ?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2309,8 +2334,25 @@ async def custom_token_exchange(
...
                     )
 
-                # Validate and return response
-                return TokenExchangeResponse(**token_data)
+                token_response = TokenExchangeResponse(**token_data)
+
+                # Surface the actor claim for delegation exchanges. Best-effort:
+                # a decode/verify hiccup must not fail an exchange the token
+                # endpoint already accepted, so act stays None on any failure.
+                if options.actor_token and token_response.id_token:
```
nandan-bhat, line 2396: One thing I noticed here: we only surface act when an actor_token was passed. But on the login path the act ends up on the session user through UserClaims (extra is set to allow), so it gets persisted no matter whether actor_token was sent. So the same exchange behaves differently depending on which method you call. Can we pick one behavior and make both paths consistent ? A short comment on the login side explaining how act reaches the user would help too, otherwise it reads like dead data.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2309,8 +2334,25 @@ async def custom_token_exchange(
...
+                    try:
+                        jwks = await self._get_jwks_cached(domain, metadata)
```
nandan-bhat, line 2398: For a delegation login we decode the id_token here for the act claim, and then login_with_custom_token_exchange decodes the same id_token again right after. JWKS is cached so it is cheap, but we are still running the full verify twice. Could we reuse the claims we already decoded here instead of verifying a second time ? Not blocking, just a small cleanup.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2309,8 +2334,25 @@ async def custom_token_exchange(
...
+                        claims = await self._verify_and_decode_jwt(
+                            token_response.id_token, jwks, audience=self._client_id
+                        )
+                        # Apply the same normalized issuer check the login path uses
+                        # before trusting any claim from the token.
+                        if self._normalize_url(claims.get("iss", "")) == self._normalize_url(metadata.get("issuer")):
+                            token_response.act = claims.get("act")
```
nandan-bhat, line 2405: Worth calling out that act here comes from the id_token, not from the access token the resource server actually uses for authorization. The two are not guaranteed to be the same. Can we add a line in CustomTokenExchange.md saying response.act is informational and sourced from the id_token ? Otherwise people might assume it reflects what gets enforced downstream.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -257,33 +257,15 @@ class CustomTokenExchangeOptions(BaseModel):
         organization: Organization identifier for the token exchange (optional)
         authorization_params: Additional OAuth parameters (optional)
     """
-    subject_token: str = Field(..., min_length=1)
-    subject_token_type: str = Field(..., min_length=1)
+    subject_token: str
```
nandan-bhat, line 265: By moving validation out of the model and removing min_length, bad tokens now fail later at call time as a CustomTokenExchangeError instead of at construction as a Pydantic ValidationError. That is fine for a beta, but it does change the exception people catch. Can we add a note in the changelog so anyone catching ValidationError knows to update ?

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -3109,9 +3110,37 @@ async def test_custom_token_exchange_missing_actor_token_type():
                 actor_token_type=None
             )
         )
+    assert exc.value.code == CustomTokenExchangeErrorCode.MISSING_ACTOR_TOKEN_TYPE
     assert "actor_token_type" in str(exc.value).lower()
 
 
+@pytest.mark.asyncio
+async def test_custom_token_exchange_missing_actor_token():
```
nandan-bhat, line 3121: Since we are keeping these validations, could we also add tests for the cases we lost ? Right now there is no test for an empty or whitespace subject_token_type, and none for a whitespace-only actor_token. Those are exactly the gaps that would catch the two regressions I flagged in server_client.py.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -3541,6 +3774,83 @@ async def test_login_with_custom_token_exchange_failure_propagates(mocker):
...
+            subject_token_type="urn:acme:mcp-token",
+            actor_token="service-token",
+            actor_token_type="urn:ietf:params:oauth:token-type:access_token",
+        )
+    )
+
+    assert result.state_data["user"]["act"] == {"sub": "agent|abc", "act": {"sub": "svc|xyz"}}
+
+
+def test_act_claim_survives_token_refresh():
```
nandan-bhat: This test checks that update_state_data keeps user.act, but update_state_data never touches the user dict, so it passes no matter what. It is an okay guard, but it does not really exercise the refresh path. Could we point it at the actual refresh_token_grant flow so it tests the thing we actually care about ?

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: We checked how the tenant responds to it and it does reject it cleanly with a 400,so it was never unsafe, just a worse experience since you'd get a generic `TOKEN_EXCHANGE_FAILED` back instead of a clear local error. I've added the same empty/whitespace check we do for subject_token, plus a test for it. 

Addressed in [2feb760](https://github.com/auth0/auth0-server-python/pull/122/commits/2feb7603a92da379d37ea4d3a2c06965ef4f6587)

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: We did verify the tenant rejects a whitespace `actor_token` on its own (it comes back as **_"Invalid actor token"_**),so there was no security gap, but failing locally with a clear message is better. Added the empty/whitespace check for `actor_token` and a test for it. 

Addressed in [2feb760](https://github.com/auth0/auth0-server-python/pull/122/commits/2feb7603a92da379d37ea4d3a2c06965ef4f6587)

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: The difference is intentional. On the login path we persist the full set of verified claims onto the session user, so anything Auth0 put in the `id_token` (including act) comes along - it's not gated on `actor_token` because login is about establishing the whole session, not just delegation. 

On the stateless path there's no session,so we only do the extra work of reading act when the caller actually asked for delegation by passing an actor_token. In practice act only exists when there was an actor_token anyway, so the two don't really diverge in real use. You're right that the login side read like dead data though, so I've added a short comment there explaining that act reaches the user through UserClaims.

Addressed in [2feb760](https://github.com/auth0/auth0-server-python/pull/122/commits/2feb7603a92da379d37ea4d3a2c06965ef4f6587)

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: This one's intentional and we'd like to keep it as is. 

The second decode on the login path isn't redundant in purpose - it's the security gate that decides whether a session gets created at all. It validates the issuer, maps each verification failure to a specific error, and builds the user object from the verified claims. The act decode up here is the opposite: it runs after the exchange already succeeded and is best-effort, so it swallows any failure and just leaves act as None. 

We don't want login to reuse that lenient behavior, because then a token that fails verification could quietly produce an empty user and still get persisted as a valid session. So they look like duplicate work but they're doing two different jobs with opposite failure handling. 

The only real cost is one extra in-memory signature check on the delegation-login path (the JWKS is cached, so no network), which we think is worth keeping the two concerns cleanly separated.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: I dug into this against the platform token code to be sure. It turns out Auth0 writes the same act claim onto both the `id_token` and the access token from the same source, so they reflect the same acting party - they don't diverge. The one real caveat is that the access token can be opaque, in which case you can't read act off it  directly, so the id_token is the practical place to read it. 

I've added a note in `CustomTokenExchange.md` saying exactly that.

Addressed in [2feb760](https://github.com/auth0/auth0-server-python/pull/122/commits/2feb7603a92da379d37ea4d3a2c06965ef4f6587)

src/auth0_server_python/auth_types/__init__.py
kishore7snehil: Agreed the exception change is worth flagging - before this, bad input failed at construction as a Pydantic `ValidationError`, now it fails at call time as  `CustomTokenExchangeError`. 

I'll call this out clearly in the PR description so it's visible.

src/auth0_server_python/tests/test_server_client.py
kishore7snehil: Done. [Added](https://github.com/auth0/auth0-server-python/pull/122/commits/2feb7603a92da379d37ea4d3a2c06965ef4f6587) tests for both - an empty/whitespace `subject_token_type` and a whitespace-only `actor_token`, each asserting the local `INVALID_TOKEN_FORMAT` error.

src/auth0_server_python/tests/test_server_client.py
kishore7snehil: Addressed in [2feb760](https://github.com/auth0/auth0-server-python/pull/122/commits/2feb7603a92da379d37ea4d3a2c06965ef4f6587)

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2287,10 +2287,45 @@ async def custom_token_exchange(
...
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "actor_token cannot be empty or whitespace-only"
+                )
+            if options.actor_token and options.actor_token.strip().startswith("Bearer "):
+                raise CustomTokenExchangeError(
+                    CustomTokenExchangeErrorCode.INVALID_TOKEN_FORMAT,
+                    "actor_token should not include 'Bearer ' prefix"
+                )
+            if options.actor_token and not options.actor_token_type:
```
rmad17, line 2318: Should we not make the check  `not options.actor_token_type.strip()` similar to other cases?

src/auth0_server_python/auth_server/server_client.py
rmad17, line 2413: If there is a Pydantic Validation error on a missing field or an incorrect type value, that will throw invalid token format. That would be misleading.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: The pairing is already covered - `actor_token` without a type throws `MISSING_ACTOR_TOKEN_TYPE`, type without a token throws `MISSING_ACTOR_TOKEN`, so they always travel together. The only case left is a real `actor_token` with a whitespace-only type, and the tenant rejects that cleanly anyway since `actor_token_type` has to be a validtoken-type URI. So I'd leave this as is - the pairing checks plus tenant validation already cover it.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: if you build CustomTokenExchangeOptions yourself and pass the object in, Pydantic raises at your own construction site and never reaches this block. It only triggers when someone passes a raw dict and we build the model internally, so it's a defensive path for a malformed request, not something a typed caller hits. The message also carries Pydantic's own detail naming the field that's missing or mistyped, so the cause isn't lost.The info a developer needs to debug is already there.

rmad17 (dismissed):
LGTM!

rmad17 (approved):

PR #139, opened 2026-07-21
feat: add Session Transfer Token support for CTE impersonation via session transfer

kishore7snehil:
### 📋 Changes

This PR adds the initiator surface for Custom Token Exchange (CTE) Impersonation via Session Transfer (RFC 8693) to auth0-server-python. An application can exchange a subject token for a short-lived, single-use Session Transfer Token (STT) that redeems into an impersonated web session on a target app - logging an agent in as a customer with the agent recorded in the `act` claim. The SDK mints the STT and builds the redirect; it never decodes or stores the token.

#### ✨ Features
- `request_session_transfer_token` performs the CTE exchange against the `urn:{domain}:session_transfer` audience, built from the request-resolved domain (MCD-compatible), and returns a `SessionTransferTokenResult`
- The STT is opaque and single-use; the exchange is stateless and never persists it to the session
- The actor defaults to the agent's session ID token, verified against JWKS and refreshed when expired, and an explicit `actor_token` overrides it
- A passed-but-blank `actor_token` is rejected with `INVALID_TOKEN_FORMAT`, and when no usable actor can be resolved the call fails client-side with `ACTOR_UNAVAILABLE` before any network request
- `build_session_transfer_redirect` returns the URL-encoded redirect that hands the STT to the target app's login URL, forwarding `organization` when present

#### 🔧 API Changes
- New method `request_session_transfer_token(subject_token, subject_token_type, actor_token=None, actor_token_type=None, scope=None, organization=None, store_options=None) -> SessionTransferTokenResult`
- New method `build_session_transfer_redirect(target_login_url, result, organization=None) -> str`
- New `SessionTransferTokenResult` model (`session_transfer_token`, `issued_token_type`, `expires_in`, `token_type`, `scope`)
- New error codes `CustomTokenExchangeErrorCode.ACTOR_UNAVAILABLE`, `SETACTOR_REQUIRED`, `SESSION_TRANSFER_DISABLED`

#### 📖 Documentation
- Added an "Impersonation via Session Transfer (STT)" section to `examples/CustomTokenExchange.md` covering the initiator and target roles, actor sourcing and override, the redirect-URL security note, and the STT error codes
- Linked the new guide from the Custom Token Exchange section of `README.md`

#### 🧪 Testing

### Contributor Checklist


src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2635,6 +2642,171 @@ async def login_with_custom_token_exchange(
...
+            organization: Organization identifier to forward (optional)
+
+        Returns:
+            A URL string with session_transfer_token (and organization) as query parameters
+        """
+        params = {"session_transfer_token": result.session_transfer_token}
+        if organization:
+            params["organization"] = organization
+
+        return URL.build_url(target_login_url, params)
```
nandan-bhat, line 2850: `build_session_transfer_redirect` builds the URL without validating `target_login_url`. The STT is a single use credential, so if this value ever comes from untrusted input it can leak the token to another host. Can we check that the URL is absolute and uses `https` here (allowing `http` only for `localhost`/`127.0.0.1`/`[::1]` for local dev)? The docstring says nothing about this, and I think we should enforce it in code rather than trust the caller. Right now anything that `URL.build_url` accepts will go through.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2635,6 +2642,171 @@ async def login_with_custom_token_exchange(
...
+                actor_token=actor_token,
+                actor_token_type=actor_token_type,
+                organization=organization,
+            )
+
+            response = await self.custom_token_exchange(options, store_options)
+
+            return SessionTransferTokenResult(
+                session_transfer_token=response.access_token,
+                issued_token_type=response.issued_token_type or SESSION_TRANSFER_TOKEN_TYPE,
```
nandan-bhat: We fall back to the hardcoded URN with `response.issued_token_type or SESSION_TRANSFER_TOKEN_TYPE`. I would avoid fabricating this value, because if the server ever returns a response without `issued_token_type` (or a non STT one), we would still label it as an STT. Since the whole point is that callers branch on `issued_token_type`, can we surface exactly what the server returned instead of defaulting it?

src/auth0_server_python/error/__init__.py
```diff
@@ -254,6 +254,9 @@ class CustomTokenExchangeErrorCode:
     MISSING_ACTOR_TOKEN = "missing_actor_token"
     TOKEN_EXCHANGE_FAILED = "token_exchange_failed"
     INVALID_RESPONSE = "invalid_response"
+    ACTOR_UNAVAILABLE = "actor_unavailable"
+    SETACTOR_REQUIRED = "setactor_required"
```
nandan-bhat, line 258: `SETACTOR_REQUIRED` and `SESSION_TRANSFER_DISABLED` are defined here and mentioned in the docs, but I could not find anywhere they are actually raised. So today a real server 400 for these cases would surface as `TOKEN_EXCHANGE_FAILED`, not the typed code, and a developer catching on these constants would never hit them. Can we map the server error to these codes where the exchange fails (or if that is intentionally deferred, add a short comment saying they are placeholders for now)?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2635,6 +2642,171 @@ async def login_with_custom_token_exchange(
...
+        Full verification against JWKS - the same path the login callback uses - so an expired
+        or tampered token is rejected client-side rather than sent to the server as a dud.
+        """
+        if not token:
+            return False
+        try:
+            domain = await self._resolve_current_domain(store_options)
+            metadata = await self._get_oidc_metadata_cached(domain)
+            jwks = await self._get_jwks_cached(domain, metadata)
+            await self._verify_and_decode_jwt(token, jwks, audience=self._client_id)
```
nandan-bhat, line 2664: `_is_id_token_usable` verifies the ID token with `audience=self._client_id`. That is correct for a standard Auth0 ID token, just flagging it as an assumption: if a tenant ever issues the session ID token with a different `aud`, this verification would fail and the token would be treated as unusable (leading to a refresh or `ACTOR_UNAVAILABLE`). Probably fine, but worth a comment so it is a conscious choice.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2635,6 +2642,171 @@ async def login_with_custom_token_exchange(
...
+            try:
+                refreshed = await self.get_token_by_refresh_token({
+                    "refresh_token": state_data["refresh_token"],
+                    "domain": state_data.get("domain") or self._domain,
+                })
+                updated_state_data = State.update_state_data(
+                    self.DEFAULT_AUDIENCE_STATE_KEY, state_data, refreshed)
+                await self._state_store.set(self._state_identifier, updated_state_data, options=store_options)
+                session_id_token = refreshed.get("id_token") or session_id_token
+            except Exception:
```
nandan-bhat: Small one: the refresh here swallows every exception with a bare `except Exception` and sets `session_id_token = None`, which then falls through to `ACTOR_UNAVAILABLE`. That is a reasonable outcome, but silently eating all errors can hide real problems (network, config, a bug in the refresh path). Could we at least log the caught error before falling back, so failures are debuggable?

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [a070d41](https://github.com/auth0/auth0-server-python/commit/a070d419debb9ec137a70de2751aeb9ded1b247b)

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [a070d41](https://github.com/auth0/auth0-server-python/commit/a070d419debb9ec137a70de2751aeb9ded1b247b)

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [a070d41](https://github.com/auth0/auth0-server-python/commit/a070d419debb9ec137a70de2751aeb9ded1b247b)

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [a070d41](https://github.com/auth0/auth0-server-python/commit/a070d419debb9ec137a70de2751aeb9ded1b247b)

src/auth0_server_python/error/__init__.py
kishore7snehil: These aren't raised by any SDK path on purpose. I checked the server (`CustomTokenExchangeProfile.js` and its `custom_token_exchange.n2w.tests.js`): all three STT rejections come back as `error: "invalid_request"` with the reason only in `error_description` (`"setActor is required when requesting a session transfer token via token exchange."`, `"Feature is disabled for this client"`, etc.). The strings `setactor_required`/`session_transfer_disabled` don't exist anywhere in the server — it never emits them as codes.

The only way to set our typed code would be to string-match that `error_description`, which is brittle (a server reword silently breaks it), so we surface the raw server `error`/`error_description` as-is — the same as our existing CTE failures and the sibling `auth0-auth-js` STT PR. Keeping the two constants as named references for the conditions; happy to drop them if you'd rather not have unused constants.

nandan-bhat (dismissed):

nandan-bhat (approved):

PR #142, opened 2026-07-28
docs: add CLAUDE.md  and AGENTS.md guidelines for auth0-server-python

rmad17:
### Changes

Adds AI-agent guidance files so Claude Code, Codex CLI, Gemini CLI, and any other tool that reads `AGENTS.md` load this repo's actual conventions instead of inferring them. Generated by the `team-sdks:generating-claude-md` skill (v1.16.0).

  | File | Lines | Contents |
  |------|-------|----------|
  | `CLAUDE.md` | 168 | Single source of truth: persona, working principles, project overview/structure, three-tier boundaries (11 Always Do / 7 Ask First / 8 Never Do), security considerations, linked pointers to
  the reference files |
  | `AGENTS.md` | 3 | Imports `@./CLAUDE.md` per the [agent-files](https://developer.android.com/studio/gemini/agent-files) spec — no duplicated content |
  | `references/code-style.md` | 124 | `.ruff.toml` rule sets, naming table, good/bad method examples, dominant patterns |
  | `references/testing.md` | 76 | pytest conventions, the `@pytest.mark.asyncio` requirement, mocking approach |
  | `references/commands.md` | 67 | Exact CI invocations, single-test runs, and what deliberately doesn't exist (no typecheck, no live-test tier) |
  | `references/pitfalls.md` | 57 | Eight repo-specific gotchas (Python 3.9 floor, MCD domain resolver, `store_options` threading, …) |
  | `references/git-workflow.md` | 55 | Branch naming, PR template requirements, required checks |
  | `references/docs-update.md` | 50 | Tracked-docs inventory, code-to-docs mapping, feature → guide map |

  **Why `references/` and not `docs/`.** `.gitignore:25` ignores `docs` as build
  output — files placed there would not be tracked.

  ### References

  - Generated by `team-sdks:generating-claude-md` v1.16.0 (Mode 1, then re-validated
    in Mode 2 with no content changes)

  To review: read `CLAUDE.md` end to end — it should contain nothing you'd disagree
  with as a maintainer — then spot-check any reference file against its source of
  truth. The boundaries encode team judgment; the Ask-First list is the place to
  push back.

  - [ ] This change adds unit test coverage — n/a, documentation only
  - [ ] This change adds integration test coverage — n/a, documentation only
  - [x] This change has been tested on the latest version of the platform/language —
        lint and the full suite pass locally; CI covers 3.9/3.10/3.11/3.12

  The one real effect is on **AI agents working in this repo**: they will follow the
  documented boundaries. An incorrect boundary would misdirect an agent, which is
  why the boundaries warrant a maintainer's read rather than a rubber stamp.

  ### Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,19 +19,9 @@ Apply these on every task in this repo — they keep changes correct, small, and
 
 ---
 
-## Project Overview
-
-**auth0-server-python** is a library for implementing user authentication in Python applications — the server-side (confidential client) SDK that framework wrappers such as `auth0-fastapi` build on.
-
-- **Language:** Python, floor `>=3.9` (CI matrix: 3.9 / 3.10 / 3.11 / 3.12)
-- **Tech Stack:** OAuth 2.0 / OIDC via Authlib, `httpx` async transport, Pydantic v2 models, JWE-encrypted store payloads (jwcrypto), DPoP (RFC 9449)
-- **Package Manager:** Poetry (CI pins `2.2.1`); published to PyPI as `auth0-server-python`
-- **Version source of truth:** `.version` (mirrored in `pyproject.toml` — keep in sync)
-- **Dependencies:** authlib `^1.2`, httpx `^0.28.1`, pydantic `^2.10.6`, jwcrypto `^1.5.7` · test: pytest + pytest-asyncio + pytest-mock. Full list in `pyproject.toml`.
-
----
-
 ## Project Structure
+
+**auth0-server-python** is the server-side (confidential client) SDK that framework wrappers such as `auth0-fastapi` build on.
 
 ```
 src/auth0_server_python/
@@ -53,19 +43,6 @@ examples/                       # 11 per-feature Markdown guides (not runnable a
 references/                     # Agent reference docs (this file's offloaded sections)
 ```
 
-### Key Files
-
-| File | Why it matters |
-|------|----------------|
-| `src/auth0_server_python/auth_server/server_client.py` | The public API surface — ~29 public methods; most changes land here |
-| `src/auth0_server_python/auth_types/__init__.py` | Every public model; changing a field is a compatibility event |
-| `src/auth0_server_python/error/__init__.py` | Typed error hierarchy integrators branch on |
-| `src/auth0_server_python/store/abstract.py` | The contract integrators implement — changes break every downstream store |
-| `.ruff.toml` | Lint config (root file, not `pyproject.toml`) — the only style gate in CI |
-| `pyproject.toml` | Deps, Python floor, pytest `addopts` (coverage flags) |
-| `.github/workflows/test.yml` | The authoritative test + lint commands and Python matrix |
-| `.version` / `.shiprc` | Release version source; `pyproject.toml` must match |
-
 ---
 
 ## Boundaries
@@ -73,7 +50,7 @@ references/                     # Agent reference docs (this file's offloaded se
 ### ✅ Always Do
 
 - Run `poetry run pytest` and `poetry run ruff check .` before committing — both gate CI.
-- Add tests for new functionality, and mark async tests `@pytest.mark.asyncio` (there is no `asyncio_mode = auto`; an unmarked async test never runs but reports as passing).
+- Add tests for new functionality, and mark async tests `@pytest.mark.asyncio` — see Common Pitfalls for why.
 - Raise a typed error from `error/` with a stable `code`, and re-raise the SDK's own errors untouched inside a catch-all — never return `None` or a bare `dict` to signal failure.
 - Return Pydantic models from `auth_types/`, validated via `model_validate` — never leak an unvalidated response `dict` through the public API.
 - Resolve the domain through `await self._resolve_current_domain(store_options)` and thread `store_options` through every new public flow method — reading `self._domain` breaks MCD deployments and cookie-backed stores.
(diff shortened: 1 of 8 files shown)


references/testing.md
```diff
@@ -0,0 +1,76 @@
...
+- `test_init_no_secret_raises`
+- `test_start_interactive_login_no_redirect_uri`
+- `test_resolver_exception_wrapped_in_domain_resolver_error`
+
+## Organization
+
+Two shapes coexist; match the file you're editing.
+
+- **Module-level functions** — `test_server_client.py` and `test_dpop_auth.py`
+- **Grouping classes** — `test_mfa_client.py`, `test_my_account_client.py` use `TestMfaClientConstructor`,
```
kishore7snehil: `test_my_account_client.py` is grouped here with `test_mfa_client.py` as using grouping classes, but it doesn't have any `Test*` classes, it's 68 module-level test functions with `# ===` banners. The two class names on the next line (`TestMfaClientConstructor`, `TestDomainResolution`) only exist in `test_mfa_client.py`.

Shall we move `test_my_account_client.py` down to the module-level-functions bullet, so `test_mfa_client.py` is the only grouping-class example?

references/git-workflow.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Branches
+
+`main` is the default branch. Observed naming in this repo (pick the shape that matches the work):
+
+| Shape | Example | Use for |
+|-------|---------|---------|
+| `<feature>-support` | `passkey-support`, `dpop-support`, `myaccount-support` | a new capability |
+| `SDK-<ticket>-<slug>` | `SDK-8833-organisations-support` | ticket-tracked work |
+| `fix/<slug>` | `fix/cte-model-validators` | bug fixes |
```
kishore7snehil: The `fix/cte-model-validators` example doesn't match any branch in this repo. The only `fix/` branches that ever existed are things like `fix/mfa-breaking-changes`. Every other example in this table (`passkey-support`, `SDK-8833-organisations-support`, `feat/cte-delegation`, `docs/sync-2026-06-30`) points at a real PR, so this one stands out.

Can we swap it for a real one, `fix/mfa-breaking-changes` (#104) is an actual bug-fix branch?

references/pitfalls.md
```diff
@@ -0,0 +1,57 @@
...
+## 3. Async tests silently pass if you forget `@pytest.mark.asyncio`
+
+`asyncio_mode` is not set to `auto` anywhere. An async test without the marker is collected, never
+awaited, and reported as passing. If a new async test passes on the first run without you making it
+pass, check for the marker first.
+
+## 4. Never construct `httpx.AsyncClient` directly
+
+`ServerClient._get_http_client()` and `MyAccountClient._get_http_client()` merge
+`self._telemetry_headers` into every request. A hand-rolled client silently drops the `Auth0-Client`
```
kishore7snehil: `MyAccountClient._get_http_client()` merges `self._headers`, not `self._telemetry_headers` (that attribute only exists on `ServerClient`). It still works because `ServerClient` builds the MyAccount client with `headers=self._telemetry_headers`, so the header does get attached, but the name here doesn't match the code an agent would open to.

references/pitfalls.md
```diff
@@ -0,0 +1,57 @@
...
+
+A 3.10-only construct passes locally on a newer interpreter and fails only the 3.9 leg.
+
+## 2. `ruff format` is not a CI gate, and the tree is not format-clean
+
+CI runs `ruff check .` only. `ruff format --check .` currently reports 13 files that would be
+reformatted, including `server_client.py`'s test file and `helpers.py`. Running `ruff format .`
+repo-wide produces a thousand-line diff unrelated to your change. Format only what you touched.
+
+## 3. Async tests silently pass if you forget `@pytest.mark.asyncio`
```
kishore7snehil: I ran this in the repo's venv (pytest 7.4.4, pytest-asyncio 0.23.8, strict mode): an unmarked async test isn't reported as passing, it's skipped with a `PytestUnhandledCoroutineWarning` and shows up as `s` / "skipped" in the summary. `testing.md` already describes it correctly as skipped; it's this file and CLAUDE.md that say "passes."

The heading and the "if a new async test passes on the first run, check for the marker" tip point an agent at the wrong symptom, it'll be looking for a pass when the real signal is a skip. Can we reword both to "skipped (with a warning), never executed"?

CLAUDE.md
```diff
@@ -0,0 +1,168 @@
...
+| `.version` / `.shiprc` | Release version source; `pyproject.toml` must match |
+
+---
+
+## Boundaries
+
+### ✅ Always Do
+
+- Run `poetry run pytest` and `poetry run ruff check .` before committing — both gate CI.
+- Add tests for new functionality, and mark async tests `@pytest.mark.asyncio` (there is no `asyncio_mode = auto`; an unmarked async test never runs but reports as passing).
```
kishore7snehil: Same thing as the async pitfall: an unmarked async test is skipped with a warning, not "reports as passing." I confirmed it by running one in the repo's venv. Shall we bring this in line with `testing.md`'s "collected and skipped, never runs"?

references/docs-update.md
```diff
@@ -0,0 +1,50 @@
...
+This repo is a **library/SDK** — its public surface is the exported classes and methods of
+`ServerClient`, `MyAccountClient`, `MfaClient`, plus the Pydantic models in `auth_types/` and the
+error classes in `error/`. Docs track that surface.
+
+## Tracked docs
+
+| Doc | What it covers | Status |
+|-----|----------------|--------|
+| `README.md` | Install, `ServerClient` construction, interactive login, custom token exchange, MCD, session expiry, passkeys, My Account, DPoP — each a short section that links out to `examples/` | present |
+| `EXAMPLES.md` | — | **❌ missing.** This repo puts per-feature guides in `examples/*.md` instead. Add a sample to the matching `examples/` guide; only create `EXAMPLES.md` if the team decides to consolidate. |
```
kishore7snehil: The red-X "❌ missing" reads like an actionable gap, even though the same cell then says the repo uses `examples/*.md` on purpose. Might be clearer as "not used by design" so an agent doesn't try to create `EXAMPLES.md`.

kishore7snehil (changes requested):

CLAUDE.md
rmad17: Yes, instead of mentioning the same thing(mark as skipped with warning) as earlier, we should ask the agent to look at Common Pitfalls for reasoning.

PR #149, opened 2026-08-04
docs: adding repo level claude.md

rmad17:
**Added**

  - doc: Added CLAUDE.md as the single source of truth for AI-agent guidance, including persona, working principles, project structure, repo boundaries, and security considerations
  - doc: Added AGENTS.md to import CLAUDE.md so Claude Code, Codex CLI, Gemini CLI, and other compatible tools load the repo conventions
  - doc: Added references/ docs for code style, testing, commands, pitfalls, git workflow, and docs-update guidance

**Changed**

  - doc: Centralized AI-agent instructions around tracked references/ files instead of untracked docs/ output
  - doc: Documented always-do, ask-first, and never-do boundaries to reduce incorrect agent assumptions in this repo


CLAUDE.md
```diff
@@ -0,0 +1,146 @@
...
+- Deprecating a public name — use the PEP 562 `__getattr__` alias pattern in `auth_types/`; don't delete it.
+
+### 🚫 Never Do
+
+- Commit secrets, API keys, tokens, or a real Auth0 tenant domain. Test fixtures use `<client_id>`-style placeholders and `auth0.local`.
+- Log, `print`, or include in an error message any token, `code`, `code_verifier`, `client_secret`, DPoP private key, or full response body. There is no logger in the SDK — don't introduce one that emits these.
+- Fail open. A validation, resolver, JWKS, or token-endpoint failure must raise, never fall through to a permissive default (no default domain, no "assume valid").
+- Validate `state` by comparing strings. `complete_interactive_login` looks the transaction up by `{transaction_identifier}:{state}` and raises `MissingTransactionError` on a miss — that store lookup *is* the binding. If you ever add a secret-to-secret comparison, use `hmac.compare_digest`, never `==`.
+- Run `ruff format .` repo-wide — it is not a CI gate and the tree is not format-clean (13 files would change). Format only lines you touched.
```
kishore7snehil: The "13 files" count is a transcribed value that will go stale the moment someone runs ruff format on one file, and nothing in CI catches it. The src tree, the full ruff rule list, the Poetry 2.2.1 pin, and the 3.9-3.12 matrix have the same problem: they copy a value that already lives in config.

Shall we point at the command that produces the fact (`poetry run ruff format --check .`) and keep only the reasoning, so these can't drift?

references/pitfalls.md
```diff
@@ -0,0 +1,58 @@
...
+
+`ServerClient._get_http_client()` merges `self._telemetry_headers`, and `MyAccountClient._get_http_client()`
+merges `self._headers`, into every request (`ServerClient` constructs its `MyAccountClient` with
+`headers=self._telemetry_headers`, so the header still reaches it). A hand-rolled client silently drops
+the `Auth0-Client` header, so the call becomes invisible to SDK telemetry. Same reasoning for auth: use
+`BearerAuth` / `DPoPAuth` from `auth_schemes/` rather than setting `Authorization` yourself.
+
+## 5. `domain` may be a callable — resolve it, don't read it
+
+MCD mode accepts `domain=<async resolver>` instead of a string. Any code path that builds a URL must
```
kishore7snehil, line 36: This MCD entry gives an agent a real recipe: resolve through `_resolve_current_domain`, thread `store_options`, fail closed. CTE, CIBA, and passkeys get no equivalent, so an agent works those flows cold.

Shall we add a short "to work on X, read these files" pointer for each complex flow instead of a full inline explanation?

CLAUDE.md
```diff
@@ -0,0 +1,146 @@
...
+- Attach credentials with `BearerAuth` / `DPoPAuth` from `auth_schemes/`, not by setting `Authorization` inline.
+- Keep `Optional[X]` / `Union[...]` typing and `target-version = "py39"`-compatible syntax — the 3.9 CI leg fails on 3.10+ constructs.
+- Update `README.md` and the matching `examples/*.md` guide in the same PR when changing the public API, configuration options, or supported integration patterns.
+- Put new code snippets in the matching `examples/*.md` guide and link to it from `README.md` — don't add another code block to `README.md`.
+- Keep `.version` and `pyproject.toml`'s `version` in sync when either is touched.
+- Preserve existing declaration order and the `# ==== SECTION ====` banners; insert new methods into the matching section. In a file without banners, group a new method/class/function with the related code it belongs to, or append it at the end — never insert it at an arbitrary position.
+
+### ⚠️ Ask First
+
+- **Any breaking change — always ask first.** Never make one on your own initiative: a renamed/removed public method, a changed signature or default, a tightened model field, a new required constructor argument, or a raised Python floor.
```
kishore7snehil, line 69: This breaking-change list misses the error-code contract. The suite matches on a typed error's `code`, so renaming or removing an existing `*ErrorCode` breaks integrators even though it isn't a signature change.

Can we add it here so an existing code is treated as public API?

references/docs-update.md
```diff
@@ -0,0 +1,57 @@
...
+When you add or change a code sample, put it in the matching `examples/*.md` guide and link to it
+from `README.md`'s numbered section for that feature — don't add a new code block to `README.md`
+itself. `README.md` already carries some inline snippets from before this rule; don't add more, and
+feel free to move one to its guide while you're touching that section.
+
+## When you change code, update these docs
+
+| When this changes | Update these docs |
+|-------------------|-------------------|
+| A public method on `ServerClient` / `MyAccountClient` / `MfaClient` — added, renamed, or removed | `README.md` if it has a numbered section for the feature, **and** the matching `examples/*.md` guide |
```
kishore7snehil, line 30: This covers updating an existing `examples/*.md` guide, but not the case where a new feature has no guide yet.

Shall we add a line for creating one, modelled on a sibling guide and linked from README.md?

kishore7snehil (changes requested):

references/docs-update.md
rmad17: My thought here would be - to add a CONVENTIONS.md for this kind of instructions and ask CLAUDE.md to always verify new code/examples with CONVENTIONS.md. Then we reduced the CLAUDE.md size and load the conventions on demand.

references/pitfalls.md
rmad17: Good idea, I have added it.

CLAUDE.md
rmad17: Yes, that is good point.

references/docs-update.md
kishore7snehil: @rmad17 Did you create one under references?

kishore7snehil (changes requested):

references/git-workflow.md
```diff
@@ -0,0 +1,55 @@
...
+`main` is the default branch. Observed naming in this repo (pick the shape that matches the work):
+
+| Shape | Example | Use for |
+|-------|---------|---------|
+| `<feature>-support` | `passkey-support`, `dpop-support`, `myaccount-support` | a new capability |
+| `SDK-<ticket>-<slug>` | `SDK-8833-organisations-support` | ticket-tracked work |
+| `fix/<slug>` | `fix/mfa-breaking-changes` (#104) | bug fixes |
+| `feat/<slug>` | `feat/cte-delegation` | scoped features |
+| `docs/<slug>` | `docs/sync-2026-06-30` | docs-only changes |
+| `release/<version>` | `release/<the version in .version>` | release prep only — cut by the release flow, not by hand |
```
kishore7snehil, line 14: We should add entries for chores, ci etc. Check this out - https://www.conventionalcommits.org/en/v1.0.0/#examples

references/git-workflow.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Commits
+
+No commitlint / husky hook exists, so nothing fails CI on message format. History is mixed:
+Conventional Commits where a ticket drove it (`feat: enforce upstream IdP session_expiry ceiling
+(IPSIE SL1) (#120)`) alongside plain imperative subjects (`Added dpop support for myaccount and
+passkeys`). **Prefer Conventional Commits** (`feat:`, `fix:`, `refactor:`, `test:`, `docs:`) for new
```
kishore7snehil, line 21: We can add chore, ci, feat!/ fix! to this list

kishore7snehil (changes requested):

references/git-workflow.md
sanchitmehtagit: Something like below is sufficient 
https://github.com/auth0/auth0-nuxt/blob/a9b69a68410d34114a701e4a6f0519f0a73209bd/references/git-workflow.md?plain=1#L7
LLM understands what are conventional commits you dont need to specify all

references/git-workflow.md
sanchitmehtagit: same as https://github.com/auth0/auth0-server-python/pull/149#discussion_r3774184451

references/docs-update.md
rmad17: Not yet, will not be part of this PR. We will finalize the conventions and then add in references,

kishore7snehil (approved):
LGTM!

PR #153, opened 2026-08-08
feat: passwordless support

rmad17:
### Changes

**Added**
  - Adds ServerClient.passwordless for embedded passwordless flows: Email OTP, SMS OTP (via /passwordless/start + the passwordless OTP token grant), and Email Magic Link (SDK-owned state/redirect_uri, transaction
  storage for callback completion).
  - Creates server-side sessions after successful OTP verification and magic-link callback completion, reusing the shared _persist_session_from_token_response path (issuer/audience/session-expiry checks already
  enforced there, not new in this PR).
  - Handles mfa_required during passwordless OTP verification; MFA can be completed into a new SDK session via mfa.verify(..., persist=True). Known limitation, not yet fixed: this MFA-initiated session-creation
  path does not run organization-claim validation, unlike every other session-creating path in the SDK. Flagging for reviewer visibility — no organization-scoped passwordless/MFA usage should ship until this is
  closed.
  - Adds typed passwordless option models (StartPasswordlessEmailOptions, StartPasswordlessSmsOptions, VerifyPasswordlessOtpOptions) and passwordless-specific typed errors (PasswordlessStartError,
  PasswordlessVerifyError), including retry_after (from Retry-After), raw non-JSON error body capture (truncated), and 429 → too_many_requests mapping.
  - Enforces E.164 validation for SMS phone numbers.
  - Adds safeguards so magic-link callers cannot override SDK-owned protocol values (redirect_uri, response_type, state); supports caller-supplied scope on magic link, re-injecting openid when omitted.
  - Removes organization support from passwordless start/verify. The field existed but always failed in practice — the server ignores it on this grant, and the SDK's own claim validation then hard-raised on the
  missing claim. Removed rather than fixed, since there was no working path to preserve. Flag if organization on passwordless was already relied on anywhere.
  - Fixes session sid sourcing for all interactive logins, not just passwordless — complete_interactive_login previously fell back to a random sid whenever user_info was absent, even when the ID token carried a sid
  claim. It now prefers the ID token's sid claim, then user_info, then random. This affects every existing interactive-login session, since OIDC back-channel logout matches by sid and previously could silently
  fail to target sessions created this way.
  - Adds Passwordless documentation (examples/Passwordless.md) and README links.
  
### Testing

This has been tested for the following flows

 | # | Flow        | MFA Configuration | Result |
  |---|-------------|--------------------|--------|
  | 1 | Email OTP   | No MFA             | ✅ Pass |
  | 2 | Email OTP   | Global MFA         | ✅ Pass |
  | 3 | Email OTP   | Step Up MFA        | ✅ Pass |
  | 4 | SMS OTP     | No MFA             | ✅ Pass |
  | 5 | SMS OTP     | Global MFA         | ✅ Pass |
  | 6 | SMS OTP     | Step Up MFA        | ✅ Pass |
  | 7 | Magic Link  | No MFA             | ✅ Pass |
  | 8 | Magic Link  | Global MFA         | ✅ Pass |
  | 9 | Magic Link  | Step Up MFA        | ✅ Pass |

- [x] This change adds integration test coverage

### Checklist


Piyush-85 (dismissed):
Reviewed it async with @rmad17, LGTM

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -610,6 +626,133 @@ async def start_interactive_login(
...
+                "session_expires_at": session_expires_at,
+            },
+        )
+
+        await self._state_store.set(
+            self._state_identifier, state_data, options=store_options
+        )
+        return state_data
+
+    async def _establish_session_from_mfa_verify_response(
```
kishore7snehil: The helpers shouldn't be here.It should be where all the helpers reside.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2862,6 +2992,15 @@ def mfa(self) -> MfaClient:
         """Access the MFA client for multi-factor authentication operations."""
         return self._mfa_client
 
+    # ============================================================================
```
kishore7snehil: API definitions of independent new feature/functions should go at last - which increases the readibility.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
     created_at: int
 
 
+# =============================================================================
```
kishore7snehil: The whole section can be added to the last.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+    email: str
+    # "code" -> email OTP; "link" -> magic link.
+    send: Literal["code", "link"] = "code"
+
+
+class StartPasswordlessSmsOptions(_StartPasswordlessBase):
+    """Options for starting an SMS passwordless (OTP) flow."""
+
+    connection: Literal["sms"] = "sms"
+    # E.164 format, e.g. "+14155550100".
```
kishore7snehil: This can be dropped as we are adding a validation explicitly

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+class VerifyPasswordlessOtpOptions(BaseModel):
+    """
+    Options for verifying a passwordless OTP and establishing a session.
+
+    Exactly one of ``email`` / ``phone_number`` must be provided and must
+    match ``connection`` (email -> email, sms -> phone_number).
+    """
+
+    # Unknown keys raise rather than being silently ignored, so a caller
+    # passing an unsupported kwarg is told, not quietly dropped.
```
kishore7snehil: Looks like it tells you why we added this line while development rather than usability

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+    model_config = ConfigDict(extra="forbid")
+
+    connection: PasswordlessConnection
+    # Public field name mirrors nextjs-auth0's `verificationCode`; sent to
+    # Auth0 as the `otp` form parameter.
```
kishore7snehil: We can't use other SDK's reference here.

kishore7snehil (changes requested):

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -610,6 +626,133 @@ async def start_interactive_login(
...
+        except jwt.InvalidTokenError as e:
+            raise MfaVerifyError(f"ID token verification failed: {str(e)}") from e
+
+        token_issuer = claims.get("iss", "")
+        if self._normalize_url(token_issuer) != self._normalize_url(origin_issuer):
+            raise MfaVerifyError(
+                "ID token issuer mismatch. Ensure your Auth0 domain is configured correctly."
+            )
+
+        user_claims = UserClaims.model_validate(claims)
```
kishore7snehil: This path creates a session but never runs the organization-claim check that the interactive login runs before it persists. A first-login MFA flow started with an organization requirement would end up with a session that was never checked for org membership, and it fails open rather than closed.

Can we run the same org validation here before persisting, or gate this path so an org-scoped login cannot complete through it?

src/auth0_server_python/tests/test_passwordless_client.py
```diff
@@ -0,0 +1,1022 @@
...
+        client = _make_client(redirect_uri=None)
+        _mock_http(client, 200, {})
+
+        with pytest.raises(MissingRequiredArgumentError):
+            await client.passwordless.start(
+                StartPasswordlessEmailOptions(email="user@example.com", send="link")
+            )
+
+    @pytest.mark.asyncio
+    async def test_caller_cannot_override_reserved_param(self):
```
kishore7snehil, line 278: This passes even if the reserved-key branch is removed, because `redirect_uri` also is not in the allowlist, so the second guard raises the same `InvalidArgumentError` and the test cannot tell which one fired.

Shall we assert on the message, or use a key that is allowed but reserved, so the test actually pins the reserved-key behavior?

src/auth0_server_python/tests/test_passwordless_client.py
```diff
@@ -0,0 +1,1022 @@
...
+            )
+        # Token is re-encrypted by the SDK, never passed through raw.
+        assert exc.value.mfa_token is not None
+        assert exc.value.mfa_token != "raw_server_mfa_token"
+        decrypted = client._mfa_client.decrypt_mfa_token(exc.value.mfa_token)
+        assert decrypted.mfa_token == "raw_server_mfa_token"
+        client._state_store.set.assert_not_awaited()
+
+    @pytest.mark.asyncio
+    async def test_passwordless_mfa_verify_persist_creates_session(self, mocker):
```
kishore7snehil, line 884: There is no test covering an org-scoped login that completes through the MFA-into-session path. Given that path skips org validation today, a test that starts with an organization and asserts the resulting behavior would make the gap visible rather than silent.

Can we add one?

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+        if not re.match(_BCP47_LANGUAGE_RE, value):
+            raise ValueError("language must be a valid BCP 47 tag (e.g. 'fr', 'en-US')")
+        return value
+
+
+class StartPasswordlessEmailOptions(_StartPasswordlessBase):
+    """Options for starting an email passwordless flow (OTP code or magic link)."""
+
+    connection: Literal["email"] = "email"
+    email: str
```
kishore7snehil: `email` is accepted as any string, so a malformed value or one with embedded CRLF passes model validation and is only rejected later by Auth0.

Shall we validate the format here the way phone numbers are validated, so a bad value fails fast?

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+    """
+
+    # Unknown keys raise rather than being silently ignored, so a caller
+    # passing an unsupported kwarg is told, not quietly dropped.
+    model_config = ConfigDict(extra="forbid")
+
+    connection: PasswordlessConnection
+    # Public field name mirrors nextjs-auth0's `verificationCode`; sent to
+    # Auth0 as the `otp` form parameter.
+    verification_code: str
```
kishore7snehil: `verification_code` accepts an empty string and arbitrarily long input.

A quick length or non-empty check here would fail fast instead of sending an empty OTP and getting back a generic `invalid_grant`.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+    email: Optional[str] = None
+    phone_number: Optional[str] = None
+    scope: Optional[str] = None
+    audience: Optional[str] = None
+    # End-user client IP, relayed to Auth0 as `auth0-forwarded-for` on the OTP
+    # token exchange so brute-force protection keys on the real user, not the
+    # app server. Honored only for confidential clients with "Trust Token
+    # Endpoint IP Header" enabled.
+    client_ip: Optional[str] = None
```
kishore7snehil: `client_ip` (and `captcha`, `scope`, `audience`) are accepted as any string with no format or length check. Lower priority than email and OTP, but worth deciding whether we validate these or intentionally pass them through untouched.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -610,6 +626,133 @@ async def start_interactive_login(
...
+    ) -> StateData:
+        """
+        Build and persist a session ``StateData`` from a token endpoint response.
+
+        Shared by the interactive-login callback and the passwordless OTP verify
+        flow so both derive the session id, enforce the IPSIE ceiling, and write
+        the state store identically.
+
+        The session ``sid`` is taken from the verified ID token claims (falling
+        back to userinfo, then a random value) so OIDC back-channel logout — which
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -610,6 +626,133 @@ async def start_interactive_login(
...
+        matches sessions by ``sid`` — can target sessions created here.
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -684,6 +690,179 @@ class MfaTokenContext(BaseModel):
...
+        "response_type",
+        "state",
+        "nonce",
+        "code_challenge",
+        "code_challenge_method",
+    }
+)
+
+# Caller-supplied authParams keys the SDK will forward. This is an allowlist
+# (Global §3: allowlists, not denylists) — any key outside it is rejected, so a
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_server/passwordless_client.py
```diff
@@ -0,0 +1,486 @@
+"""
+Passwordless Client for auth0-server-python SDK.
+
+Implements embedded passwordless login (Legacy Passwordless connections) for a
+confidential (Regular Web App) client:
+
+* Email OTP / SMS OTP — ``start()`` sends a code, ``verify()`` exchanges it for
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_server/passwordless_client.py
```diff
@@ -0,0 +1,486 @@
...
+  tokens via the passwordless-OTP grant and establishes a server-side session.
+* Magic link — ``start(send="link")`` emails a one-click link; completion is
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_server/passwordless_client.py
```diff
@@ -0,0 +1,486 @@
...
+        store_options: Optional[dict[str, Any]] = None,
+    ) -> PasswordlessStartResult:
+        """
+        Start a passwordless flow by sending an OTP code or a magic link.
+
+        Args:
+            options: ``StartPasswordlessEmailOptions`` or
+                ``StartPasswordlessSmsOptions``.
+            store_options: Options passed to the transaction store (e.g.
+                request/response) — required for the magic-link flow so the
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_server/passwordless_client.py
```diff
@@ -0,0 +1,486 @@
...
+        origin_domain: str,
+        store_options: Optional[dict[str, Any]],
+    ) -> tuple[dict[str, Any], tuple[str, TransactionData]]:
+        """
+        Build the magic-link ``authParams`` and the transaction to persist.
+
+        The SDK owns ``redirect_uri`` / ``response_type`` / ``state``; caller
+        ``auth_params`` may only contribute non-reserved passthrough keys.
+        Persisting the transaction is the caller's job, deferred until after
+        ``POST /passwordless/start`` succeeds — a failed start must not leave a
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/auth_server/passwordless_client.py
```diff
@@ -0,0 +1,486 @@
...
+        redirect_uri = client._redirect_uri
+        if not redirect_uri:
+            raise MissingRequiredArgumentError("redirect_uri")
+
+        auth_params = self._sanitize_caller_auth_params(options.auth_params)
+
+        if store_options is None:
+            raise MissingRequiredArgumentError("store_options")
+
+        # Auth0 echoes `state` back unvalidated on this flow — it does not
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+Passwordless has two shapes:
+
+1. **OTP code** - `start()` sends a code by email or SMS. Your app collects that code, then `verify()` exchanges it at `/oauth/token` with the passwordless OTP grant and **creates a server-side session**.
+2. **Magic link** - `start(send="link")` sends a one-click email link. Auth0 redirects the user back to your callback URL, and your app completes the flow with `complete_interactive_login()`. The callback creates the server-side session.
+
+OTP start does **not** create a session. The session exists only after `verify()` succeeds. Magic-link start writes a transaction so the callback can validate the returned `state`; the session exists only after the callback completes.
+
+## Prerequisites
+
+These flows require a **Regular Web Application** — passwordless token exchange
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+needs a client secret, which a public/SPA client cannot hold safely.
+
+Two tenant-level settings are also required and are easy to miss because
+neither failure mode looks like a configuration problem:
+
+1. **Authentication Profile must be "Identifier First."** The default
+   "Universal Login" profile blocks the direct `/oauth/token` call this SDK
+   uses for OTP verification — without it, OTP `verify()` fails with
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+   `unauthorized_client`. Set it under your tenant's Authentication Profile
+   settings. (Skip this if you only use passwordless via Universal Login
+   redirects rather than this SDK's embedded flow.)
+2. **Enable the Passwordless OTP grant type** on your application
+   (**Applications -> Your App -> Advanced Settings -> Grant Types**). Without
+   it, OTP verification also fails with `unauthorized_client`.
+3. **Magic link only** — set the tenant flag
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+   `universal_login.passwordless.allow_magiclink_verify_without_session` to
+   `true` via the Management API:
+
+   ```
+   PATCH /api/v2/tenants/settings
+   { "universal_login": { "passwordless": { "allow_magiclink_verify_without_session": true } } }
+   ```
+
+   This is required for **any** server-side SDK completing magic link (this
+   one, Express, Next.js, etc.) — the browser that opens the emailed link is
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+    store_options={"request": request, "response": response},
+)
+
+user = result["state_data"]["user"]
+```
+
+> [!WARNING]
+> Do not let callers override `redirect_uri`, `state`, `response_type`, `nonce`, or PKCE fields in magic-link `auth_params`. The SDK owns these values so the emailed authorization code and state cannot be redirected to an attacker-controlled URL.
+>
+> This matters more than it looks: Auth0 treats magic-link `state` as a pure echo and does not validate it server-side, and the clicked link's query string can overwrite whatever the browser originally stored. The SDK's single-use, `state`-keyed transaction plus the exact-match `redirect_uri` are therefore the *entire* CSRF / authorization-code-interception defense on this flow — Auth0 will not catch a bypass for you.
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+
+> [!NOTE]
+> If the callback fails (expired link, JWKS unavailable, a rejected ID token), have the user restart the flow from `start()` rather than retrying the same link — a failed callback does not guarantee the transaction was cleaned up, so re-submitting the same callback URL can produce a confusing error instead of a clear "session expired, please try again."
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+            "audience": "https://api.example.com",
+            "scope": "openid profile email offline_access read:orders",
+            "login_hint": "user@example.com",
+        },
+    ),
+    store_options={"request": request, "response": response},
+)
+```
+
+A caller-supplied magic-link `scope` **replaces** the default wholesale rather than merging with it. The SDK re-injects `openid` when your scope omits it, because the magic-link callback would otherwise complete on a token response carrying no ID token — a session built from claims nothing verified. `openid` is never duplicated and your scope order is preserved otherwise.
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+
+`PasswordlessStartError` and `PasswordlessVerifyError` both carry:
+
+- `code` / `message` - the Auth0 `error` and `error_description`, or an SDK-side code when the response had neither
+- `error` / `error_description` - the raw values from a JSON error body
+- `retry_after` - seconds from the `Retry-After` response header, typically on a 429. `None` when the header is absent or in HTTP-date form (which the SDK does not interpret)
+- `cause` - the parsed JSON error body, or the response text truncated to 2048 characters when the body was not JSON
+
+> [!WARNING]
+> `cause` may hold a raw upstream body (an HTML error page, WAF block page, or proxy dump). It is length-capped, but not redacted — do not log it at a level where untrusted upstream content is unwelcome.
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/tests/test_passwordless_client.py
```diff
@@ -0,0 +1,1022 @@
+"""
+Tests for PasswordlessClient — embedded passwordless (OTP + magic link).
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

src/auth0_server_python/tests/test_passwordless_client.py
```diff
@@ -0,0 +1,1022 @@
...
+
+# ── Magic link callback completion (complete_interactive_login) ──────────────
+
+
+class TestMagicLinkCallback:
+    @pytest.mark.asyncio
+    async def test_magic_link_callback_exchanges_code_without_pkce(self, mocker):
+        # Magic link is a plain auth-code exchange: lock that code_verifier=None
+        # reaches fetch_token (authlib drops the falsy field) so a forced verifier
+        # — which Auth0 would reject — is caught.
```
kishore7snehil: Let's use a plain hyphen here instead of the em dash, to match the repo convention.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
+# Passwordless Authentication
+
+Passwordless lets users sign in with a one-time code sent by email or SMS, or with a magic link sent by email. This guide covers the **embedded login** flow on `ServerClient.passwordless` and how each path establishes a server-side session.
+
+> [!NOTE]
+> Passwordless API flows use Auth0 Legacy Passwordless connections (`email` and `sms`). Enable the **Passwordless OTP** grant for your application under **Applications -> Your App -> Advanced Settings -> Grant Types**. See the [Auth0 Passwordless API documentation](https://auth0.com/docs/authenticate/passwordless/implement-login/embedded-login/relevant-api-endpoints).
+
+> [!IMPORTANT]
+> These flows are for confidential server-side applications. Tokens stay on the server; the browser should only receive your application's session cookie or opaque session reference.
```
kishore7snehil: Two independent clauses spliced with a semicolon here. Can we split into two sentences?

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+- [Error Handling](#error-handling)
+
+## How the flow works
+
+Passwordless has two shapes:
+
+1. **OTP code** - `start()` sends a code by email or SMS. Your app collects that code, then `verify()` exchanges it at `/oauth/token` with the passwordless OTP grant and **creates a server-side session**.
+2. **Magic link** - `start(send="link")` sends a one-click email link. Auth0 redirects the user back to your callback URL, and your app completes the flow with `complete_interactive_login()`. The callback creates the server-side session.
+
+OTP start does **not** create a session. The session exists only after `verify()` succeeds. Magic-link start writes a transaction so the callback can validate the returned `state`; the session exists only after the callback completes.
```
kishore7snehil: Same here, let's split the semicolon into two sentences.

examples/Passwordless.md
```diff
@@ -0,0 +1,383 @@
...
+    VerifyPasswordlessOtpOptions(
+        connection="sms",
+        phone_number="+14155550100",
+        verification_code=user_entered_code,
+    ),
+    store_options={"request": request, "response": response},
+)
+```
+
+By default, email OTP requests `openid profile email`; SMS OTP requests `openid profile` because SMS identities do not have an email claim to satisfy.
```
kishore7snehil: Same, two sentences instead of the semicolon.

kishore7snehil (changes requested):

src/auth0_server_python/tests/test_passwordless_client.py
rmad17: Yes, I will add one.
>Note: Org support is not there in passwordless

src/auth0_server_python/tests/test_passwordless_client.py
rmad17: Valid point, added the assertion.

src/auth0_server_python/tests/test_passwordless_client.py
rmad17: Addressed the tests here -  https://github.com/auth0/auth0-server-python/pull/153/changes/273844ddc04060ff56dae35c51179807dcbda03b

src/auth0_server_python/tests/test_passwordless_client.py
rmad17: Addressed the test here: https://github.com/auth0/auth0-server-python/pull/153/changes/273844ddc04060ff56dae35c51179807dcbda03b

kishore7snehil (approved):
LGTM!

PR #154, opened 2026-08-10
feat!: add Private Key JWT (private_key_jwt) client authentication

kishore7snehil:
### 📋 Changes

This PR adds Private Key JWT (`private_key_jwt`) client authentication to auth0-server-python, alongside the existing client secret. When a signing key is configured, the SDK authenticates to the token endpoint by signing a short-lived client assertion (RFC 7523) instead of sending a client secret. Client authentication is resolved in one place and applied uniformly across every client-authenticated call site.

#### ✨ Features
- New `client_assertion_signing_key` and `client_assertion_signing_alg` options on `ServerClient` select `private_key_jwt`; when neither a signing key nor a client secret is configured, token requests raise `ConfigurationError` before any network call
- `private_key_jwt` is applied uniformly across all client-authenticated requests: authorization code exchange, refresh token, Custom Token Exchange, connection access token (Token Vault), backchannel (CIBA) initiate and grant, PAR, passkey signin, and MFA challenge and verify
- The signing key and algorithm are validated when the client is constructed, so a malformed key or a key that does not match the algorithm fails early rather than on the first token request
- Caller-supplied `client_secret`, `client_assertion`, and `client_assertion_type` are stripped from every client-authenticated request, so the SDK-controlled credentials cannot be overridden through `authorization_params`

#### 🔧 API Changes
- New `ServerClient` option `client_assertion_signing_key: Optional[str]` - a PKCS8 PEM private key that enables `private_key_jwt`
- New `ServerClient` option `client_assertion_signing_alg: Optional[str]` - the assertion signing algorithm, defaulting to `RS256`
- New `build_client_assertion(private_key, client_id, issuer, alg)` helper, `validate_client_assertion_key(private_key, alg)` helper, and `CLIENT_ASSERTION_TYPE` constant in `auth0_server_python.auth_schemes.client_assertion`

#### ⚠️ Breaking Changes
- Client-authenticated requests now require client authentication to be configured. A client with neither a client secret nor a signing key previously completed the interactive login code exchange as a public client, and now raises `ConfigurationError`. The MFA challenge (`/mfa/challenge`) previously sent `client_secret: null` in the body for such a client, and now raises `ConfigurationError` too. Configure `client_secret` or `client_assertion_signing_key`.
- Pushed Authorization Requests now set `response_type` to `code`. This was previously missing and caused the request to be rejected. A caller that set a different `response_type` through `authorization_params` on the PAR path will have that value replaced with `code`.

#### 📖 Documentation
- Added an "Authenticating with Private Key JWT" section to the client setup in `README.md`, covering the signing-key configuration, where to register the public key, the algorithm requirement, the passkey challenge limitation, and that private keys must not be committed to source control

#### 🧪 Testing

### Contributor Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,10 @@ auth0 = ServerClient(
 )
 ```
 
-The key must be a PKCS8 PEM private key whose public key is registered on your Auth0 application. The signing algorithm defaults to `RS256` and can be overridden with `client_assertion_signing_alg`.
+The key must be a PKCS8 PEM private key. Register its public key on your Auth0 application under **Settings → Credentials**, and set the application's authentication method to Private Key JWT. The signing algorithm defaults to `RS256` and can be overridden with `client_assertion_signing_alg`. The algorithm must match the key type and the algorithm chosen when the public key credential was created.
+
+> [!NOTE]
+> The passkey challenge endpoints (`register` and `challenge`) accept only a client secret, so a client configured with just a signing key cannot use them.
 
 > [!IMPORTANT]
 > Private keys must not be committed to source control. Load them from a secure secret store or an environment-provided file.
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ auth0 = ServerClient(
 )
 ```
 
-The key must be a PKCS8 PEM private key. Register its public key on your Auth0 application under **Settings → Credentials**, and set the application's authentication method to Private Key JWT. The signing algorithm defaults to `RS256` and can be overridden with `client_assertion_signing_alg`. The algorithm must match the key type and the algorithm chosen when the public key credential was created.
+The key must be a PKCS8 PEM private key. Register its public key on your Auth0 application under **Settings → Credentials**, and set the application's authentication method to Private Key JWT. The signing algorithm defaults to `RS256` and can be overridden with `client_assertion_signing_alg`. Auth0 accepts `RS256`, `RS384`, and `PS256`, all of which use an RSA key. The algorithm must match the key type and the algorithm chosen when the public key credential was created.
 
 > [!NOTE]
 > The passkey challenge endpoints (`register` and `challenge`) accept only a client secret, so a client configured with just a signing key cannot use them.
--- a/README.md
+++ b/README.md
@@ -55,6 +55,32 @@ The `AUTH0_SECRET` is the key used to encrypt the session and transaction cookie
 openssl rand -hex 64
 ```
 
+#### Authenticating with Private Key JWT
+
+The SDK authenticates to Auth0 with either a client secret or a Private Key JWT (`private_key_jwt`). To use Private Key JWT, pass your private signing key as `client_assertion_signing_key` instead of `client_secret`:
+
+```python
+from auth0_server_python.auth_server.server_client import ServerClient
+
+with open('private_key.pem') as f:
+    private_key = f.read()
+
+auth0 = ServerClient(
+    domain='<AUTH0_DOMAIN>',
+    client_id='<AUTH0_CLIENT_ID>',
+    client_assertion_signing_key=private_key,
+    secret='<AUTH0_SECRET>',
+    authorization_params={
+        'redirect_uri': '<AUTH0_REDIRECT_URI>',
+    }
+)
+```
+
+The key must be a PKCS8 PEM private key whose public key is registered on your Auth0 application. The signing algorithm defaults to `RS256` and can be overridden with `client_assertion_signing_alg`.
+
+> [!IMPORTANT]
+> Private keys must not be committed to source control. Load them from a secure secret store or an environment-provided file.
+
 ### 3. Add login to your Application (interactive)
 
 Before using redirect-based login, ensure the `redirect_uri` is configured when initializing the SDK:
--- a/src/auth0_server_python/auth_schemes/client_assertion.py
+++ b/src/auth0_server_python/auth_schemes/client_assertion.py
@@ -29,7 +29,7 @@ def validate_client_assertion_key(private_key: Union[str, bytes], alg: str = "RS
     except Exception as e:
         raise ConfigurationError(
             f"Invalid client_assertion_signing_key for algorithm {alg}: {e}"
-        )
+        ) from e
 
 
 def build_client_assertion(
--- a/src/auth0_server_python/auth_schemes/client_assertion.py
+++ b/src/auth0_server_python/auth_schemes/client_assertion.py
@@ -1,0 +1,42 @@
+import secrets
+import time
+from typing import Union
+
+import jwt
+
+# RFC 7523 client-assertion type for private_key_jwt authentication.
+CLIENT_ASSERTION_TYPE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+
+# Assertion lifetime in seconds.
+_ASSERTION_LIFETIME = 60
+
+
+def build_client_assertion(
+    private_key: Union[str, bytes],
+    client_id: str,
+    issuer: str,
+    alg: str = "RS256",
+) -> str:
+    """
+    Mint a private_key_jwt client assertion for token-endpoint authentication (RFC 7523).
+
+    Args:
+        private_key: The client's private signing key (PKCS8 PEM string or bytes).
+        client_id: The Auth0 client ID, used as both the issuer and subject claim.
+        issuer: The authorization server issuer identifier, used as the audience claim.
+        alg: The signing algorithm (defaults to "RS256").
+
+    Returns:
+        The signed client-assertion JWT.
+    """
+    now = int(time.time())
+    payload = {
+        "iss": client_id,
+        "sub": client_id,
+        "aud": issuer,
+        "iat": now,
+        "nbf": now,
+        "exp": now + _ASSERTION_LIFETIME,
+        "jti": secrets.token_urlsafe(32),
+    }
+    return jwt.encode(payload, private_key, algorithm=alg)
--- a/src/auth0_server_python/auth_schemes/client_assertion.py
+++ b/src/auth0_server_python/auth_schemes/client_assertion.py
@@ -4,11 +4,32 @@ from typing import Union
 
 import jwt
 
+from auth0_server_python.error import ConfigurationError
+
 # RFC 7523 client-assertion type for private_key_jwt authentication.
 CLIENT_ASSERTION_TYPE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
 
 # Assertion lifetime in seconds.
 _ASSERTION_LIFETIME = 60
+
+
+def validate_client_assertion_key(private_key: Union[str, bytes], alg: str = "RS256") -> None:
+    """
+    Verify the signing key and algorithm can produce a client assertion.
+
+    Args:
+        private_key: The client's private signing key (PKCS8 PEM string or bytes).
+        alg: The signing algorithm (defaults to "RS256").
+
+    Raises:
+        ConfigurationError: If the key is malformed or does not match the algorithm.
+    """
+    try:
+        jwt.encode({"probe": True}, private_key, algorithm=alg)
+    except Exception as e:
+        raise ConfigurationError(
+            f"Invalid client_assertion_signing_key for algorithm {alg}: {e}"
+        )
 
 
 def build_client_assertion(
--- a/src/auth0_server_python/auth_server/server_client.py
+++ b/src/auth0_server_python/auth_server/server_client.py
@@ -22,6 +22,7 @@ from pydantic import ValidationError
 from auth0_server_python.auth_schemes.client_assertion import (
     CLIENT_ASSERTION_TYPE,
     build_client_assertion,
+    validate_client_assertion_key,
 )
 from auth0_server_python.auth_schemes.dpop_auth import make_dpop_proof_for_token_endpoint
 from auth0_server_python.auth_server.mfa_client import MfaClient
@@ -181,6 +182,10 @@ class ServerClient(Generic[TStoreOptions]):
         self._client_secret = client_secret
         self._client_assertion_signing_key = client_assertion_signing_key
         self._client_assertion_signing_alg = client_assertion_signing_alg or "RS256"
+        if client_assertion_signing_key:
+            validate_client_assertion_key(
+                client_assertion_signing_key, self._client_assertion_signing_alg
+            )
         self._redirect_uri = redirect_uri
         self._secret = secret
         self._default_authorization_params = authorization_params or {}
@@ -197,10 +202,10 @@ class ServerClient(Generic[TStoreOptions]):
         self._telemetry = Telemetry.default()
         self._telemetry_headers = self._telemetry.headers
 
-        # Initialize OAuth client
+        # A signing key takes precedence, so the secret is withheld to keep client auth single.
         self._oauth = AsyncOAuth2Client(
             client_id=client_id,
-            client_secret=client_secret,
+            client_secret=None if client_assertion_signing_key else client_secret,
             headers=self._telemetry_headers,
         )
 
@@ -222,6 +227,7 @@ class ServerClient(Generic[TStoreOptions]):
             state_store=self._state_store,
             state_identifier=self._state_identifier,
             headers=self._telemetry_headers,
+            apply_client_authentication=self._apply_client_authentication,
         )
 
     def _get_http_client(self, **kwargs) -> httpx.AsyncClient:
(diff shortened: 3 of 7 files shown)


src/auth0_server_python/auth_server/server_client.py
```diff
@@ -171,6 +179,8 @@ def __init__(
 
         self._client_id = client_id
         self._client_secret = client_secret
+        self._client_assertion_signing_key = client_assertion_signing_key
+        self._client_assertion_signing_alg = client_assertion_signing_alg or "RS256"
```
nandan-bhat, line 184: We keep the signing key here, but `MfaClient` is still created below with only `client_secret`, so the MFA flows never see it. That leaves two gaps:

1. `mfa_client.verify()` posts to `/oauth/token` with `client_secret` in the body. When only a signing key is set, `self._client_secret` is `None`, and httpx encodes that as `client_secret=` with an empty value. So the request goes out with no usable client auth. This is a token endpoint call, so `private_key_jwt` is not applied everywhere yet.
2. `/mfa/challenge` sends `"client_secret": null` in the JSON body for the same reason.

The result is that a client using `private_key_jwt` cannot complete MFA step up, including the `mfa_required` path we already touch in `get_token_by_refresh_token`.

Auth0 supports `private_key_jwt` on both `/oauth/token` and `/mfa/challenge`, and `auth0-auth-js` runs MFA verify through its shared client auth for this reason. Can we pass the signing key (or a bound `_apply_client_authentication`) into `MfaClient` and apply it with `in_body=True` ?

Passing the resolver instead of the raw secret would also stop this from happening again the next time we add a sub client.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -663,13 +714,21 @@ async def complete_interactive_login(
         # Exchange the code for tokens
         # Use redirect_uri from transaction if available, otherwise fall back to default
         token_redirect_uri = transaction_data.redirect_uri or self._redirect_uri
+
+        # client_secret is applied by self._oauth. private_key_jwt is passed to fetch_token as params.
+        client_auth_params: dict = {}
+        self._apply_client_authentication(
```
nandan-bhat, line 735: When both `client_secret` and `client_assertion_signing_key` are set, this path sends two client auth methods in the same request.

`self._oauth` is built with `client_secret`, so authlib picks `client_secret_basic` and adds the `Authorization` header itself. We then add `client_assertion` to the body here. I ran this against authlib 1.6.9 and httpx 0.28.1 and the request goes out with both `Authorization: Basic ...` and `client_assertion` in the body.

Either Auth0 accepts the header and we quietly keep authenticating with the secret the user is trying to move away from, or the request is rejected. Both are bad, and this is the login flow. Every other call site is fine because it returns `None` and suppresses basic auth. This one cannot, because authlib owns the header.

The realistic way to hit it is a migration where `AUTH0_CLIENT_SECRET` is still set in the environment while the signing key gets introduced. Refresh, CIBA and token exchange keep working, only login breaks, so it is a confusing one to debug.

`auth0-auth-js` resolves a single client auth where the key wins over the secret. Could we either reject both being set in `__init__`, or build `self._oauth` with `client_secret=None` when a signing key is present ? I checked the second option and it produces a clean request.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -219,6 +229,43 @@ def _get_http_client(self, **kwargs) -> httpx.AsyncClient:
...
+            params["client_assertion_type"] = CLIENT_ASSERTION_TYPE
+            return None
+
+        if self._client_secret:
+            if in_body:
+                params["client_secret"] = self._client_secret
+                return None
+            return (self._client_id, self._client_secret)
+
+        raise ConfigurationError(
```
nandan-bhat, line 274: This one is worth calling out in the PR description and the release notes.

Before this change, a client with no `client_secret` could still complete the code exchange, because authlib fell back to `token_endpoint_auth_method='none'` and posted `client_id` in the body. I confirmed that against authlib 1.6.9. After this change the same setup raises `ConfigurationError`.

I think requiring client auth is the right call, and it matches `auth0-auth-js`, which throws `MissingClientAuthError` in the same spot. The impact is small too, since the direct httpx paths already failed for those clients (`auth=(client_id, None)` makes httpx raise a `TypeError`). But the PR says there are no breaking changes, and this one is a real behaviour change on the main login path.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -571,12 +618,16 @@ async def start_interactive_login(
                     "configuration_error", "PAR is enabled but pushed_authorization_request_endpoint is missing in metadata")
 
             auth_params["client_id"] = self._client_id
+            # authlib supplies response_type on the non-PAR path, but the PAR post is built by hand.
+            auth_params["response_type"] = "code"
```
nandan-bhat, line 634: Good catch, PAR really was broken. On `main` the POST had no `response_type` at all, and `final_url` below reads `auth_params['response_type']`, which raises `KeyError` unless the app happened to set it in the constructor level `authorization_params`. There is no PAR test on `main` either, which is why this went unnoticed.

One side effect worth a release note: the apps that did set `response_type` in `authorization_params` are exactly the ones where PAR used to work, and their value is now silently replaced with `code`. Forcing `code` is fine by me since the SDK only implements the code flow, it just should not be silent.

This is also an unrelated fix riding along in a client auth PR. It might be easier to track as its own change.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -2426,17 +2501,23 @@ async def custom_token_exchange(
             # Merge additional authorization params
             if options.authorization_params:
                 # Prevent override of critical parameters
-                forbidden_params = {"grant_type", "client_id", "subject_token", "subject_token_type"}
+                forbidden_params = {
+                    "grant_type", "client_id", "subject_token", "subject_token_type",
+                    "client_assertion", "client_assertion_type",
```
nandan-bhat, line 2521: Good addition, and the test for it is meaningful, since with `client_secret` configured `_apply_client_authentication` does not touch `params` and a caller supplied key would otherwise survive into the body.

The same gap is still open in the two other places that merge caller supplied params. The PAR merge in `start_interactive_login` only filters `INTERNAL_AUTHORIZE_PARAMS`, and `initiate_backchannel_authentication` does a plain `params.update(authorization_params)`. So a caller can put `client_assertion` into the PAR or `bc-authorize` body next to our basic auth, which is the same mixed auth request as above.

Can we apply the same denylist in both places ? Adding `client_secret` to it would help as well.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -219,6 +229,43 @@ def _get_http_client(self, **kwargs) -> httpx.AsyncClient:
...
         return httpx.AsyncClient(headers=headers, **kwargs)
 
+    def _apply_client_authentication(
+        self, params: dict, issuer: str, in_body: bool = False
+    ) -> Optional[tuple[str, str]]:
+        """
+        Apply client authentication to an outgoing token request.
+
+        Args:
+            params: The outgoing token request body, mutated in place when a client assertion is injected.
```
nandan-bhat: Nit: the docstring says `params` is mutated when a client assertion is injected, but it is also mutated when `in_body=True` with a secret. Worth rewording so both side effects are obvious to the next reader.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -1625,7 +1693,7 @@ async def backchannel_authentication_grant(
                 return token_response
 
         except Exception as e:
-            if isinstance(e, (ApiError, PollingApiError)):
+            if isinstance(e, (ApiError, PollingApiError, ConfigurationError)):
```
nandan-bhat, line 1711: Nit: this is unreachable through `backchannel_authentication`, because `initiate_backchannel_authentication` runs before the try block and raises `ConfigurationError` first, and the catch all below would convert it to an `ApiError` anyway. It is still correct for anyone calling `backchannel_authentication_grant` directly, so no change needed if that was the intent.

src/auth0_server_python/auth_schemes/client_assertion.py
```diff
@@ -0,0 +1,42 @@
...
+    payload = {
+        "iss": client_id,
+        "sub": client_id,
+        "aud": issuer,
+        "iat": now,
+        "nbf": now,
+        "exp": now + _ASSERTION_LIFETIME,
+        "jti": secrets.token_urlsafe(32),
+    }
+    return jwt.encode(payload, private_key, algorithm=alg)
```
nandan-bhat, line 63: Nothing validates the key or the algorithm before we reach this line, and the errors coming out of `jwt.encode` are not wrapped. I tried a few bad configs against PyJWT 2.10.0:

- EC key left on the default `RS256`: `TypeError: ECPrivateKey.sign() takes 2 positional arguments but 3 were given`
- malformed PEM: `InvalidKeyError: Could not parse the provided public key.`
- unknown alg: `NotImplementedError: Algorithm not supported`

Called from `get_token_by_refresh_token`, all three surface to the user as "The access token has expired and there was an error while trying to refresh it", which points them at completely the wrong thing. It also only fails on the first token request rather than at construction.

`auth0-auth-js` imports the key once with `importPKCS8(key, alg)` when it builds the client auth, so a bad key or a key and alg mismatch fails early with a clear message. Could we load the key in `__init__` and wrap signing failures in `ConfigurationError` ?

For what it is worth, the claim set itself checks out against the Auth0 docs: `aud` keeps the trailing slash, the 60 second lifetime matches the documented recommendation, and `jti` comes out at 43 characters against the 64 character limit.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -2370,6 +2514,152 @@ async def test_get_token_by_refresh_token_mfa_required_raises_api_error_with_raw
     assert exc.value.mfa_requirements is None
 
 
+# =============================================================================
+# Private Key JWT (client assertion) Client Authentication
```
nandan-bhat, line 2595: The new tests cover a lot, but nothing calls `complete_interactive_login`, and the code exchange is the first flow listed in the PR description.

It is also the only path where client auth is handed to authlib instead of built by us, so it quietly depends on authlib internals: our kwargs reaching the body through `prepare_token_request`, and the auth method being chosen from whether `client_secret` is set. If an authlib upgrade stopped forwarding the assertion, this suite would stay green.

A test here would also have caught the double client auth problem I flagged in `complete_interactive_login`.

One related thing: `assert kwargs["auth"] is None` only checks the argument we pass, not what actually goes on the wire. Asserting on the request itself with `httpx.MockTransport` or `pytest-httpx` would also catch a stray `Authorization` header.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -2370,6 +2514,152 @@ async def test_get_token_by_refresh_token_mfa_required_raises_api_error_with_raw
...
+# =============================================================================
+
+def _generate_rsa_private_key_pem() -> str:
```
nandan-bhat, line 2598: Nit: this helper is defined here but first used around line 148, roughly 2400 lines further up. Moving it near the top with the other helpers would make the new PAR test easier to follow.

README.md
```diff
@@ -55,6 +55,32 @@ The `AUTH0_SECRET` is the key used to encrypt the session and transaction cookie
...
+    client_id='<AUTH0_CLIENT_ID>',
+    client_assertion_signing_key=private_key,
+    secret='<AUTH0_SECRET>',
+    authorization_params={
+        'redirect_uri': '<AUTH0_REDIRECT_URI>',
+    }
+)
+```
+
+The key must be a PKCS8 PEM private key whose public key is registered on your Auth0 application. The signing algorithm defaults to `RS256` and can be overridden with `client_assertion_signing_alg`.
```
nandan-bhat: Three things worth adding to this section:

1. "instead of `client_secret`" does not hold for every flow. `/passkey/register` and `/passkey/challenge` accept `client_secret` only and reject `client_assertion`, so a client with just a signing key cannot use the passkey challenge endpoints. `auth0-auth-js` has the same limitation and calls it out in code. A short caveat here and in `examples/Passkeys.md` would save people some digging. Same for MFA if we do not fix the `MfaClient` gap in this PR.
2. Where to register the public key in the dashboard (Application, then Credentials). That is usually the first thing people get stuck on.
3. Which values `client_assertion_signing_alg` accepts, and that the key type has to match it. Auth0 also requires `alg` to match the algorithm picked when the credential was created.

Also, `build_client_assertion` and `CLIENT_ASSERTION_TYPE` are listed as API changes in the PR description but are not documented anywhere. Should they be public, or renamed with a leading underscore ?

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13). The resolver is passed into `MfaClient` and applied with `in_body=True` on both challenge and verify.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13). The OAuth client is built with `client_secret=None` when a signing key is present, so only the assertion goes out.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Added a Breaking Changes section to the PR description. A client with neither a secret nor a signing key previously completed the code exchange as a public client and now raises `ConfigurationError`.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Kept the PAR fix in this PR since it shares the same call site, and documented the `response_type` behavior in the Breaking Changes section: it was missing before and is now set to `code`, replacing any value passed through `authorization_params` on the PAR path.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13). The resolver now strips `client_secret`, `client_assertion`, and `client_assertion_type` from every client-authenticated request, so PAR and backchannel are covered too.

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13).

src/auth0_server_python/auth_server/server_client.py
kishore7snehil: This is intentional, so no change. It is unreachable through `backchannel_authentication` but correct for a direct `backchannel_authentication_grant` call.

src/auth0_server_python/auth_schemes/client_assertion.py
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13). The key and algorithm are validated in `__init__` and raise `ConfigurationError`.

src/auth0_server_python/tests/test_server_client.py
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13). Added a test that inspects the outgoing request through a mock transport and asserts no basic-auth header alongside the assertion.

src/auth0_server_python/tests/test_server_client.py
kishore7snehil: The MFA tests are co-located in their factor classes now. The key-generation helpers in this file are shared across many tests, so I left them where the other module helpers sit rather than move them up.

README.md
kishore7snehil: Fixed in [c1a897f](https://github.com/auth0/auth0-server-python/commit/c1a897f3a3260cef93769ee609d11c5bf6108f13). Added the public-key registration step, the algorithm requirement, and the passkey challenge limitation. Kept `build_client_assertion` and `CLIENT_ASSERTION_TYPE` public as building blocks and documented them in the PR's API Changes.

nandan-bhat (approved):

PR #156, opened 2026-08-14
feat: add support for anonymous sessions

rmad17:
### Changes

**Added**

- Adds ServerClient.anonymous for pre-login anonymous sessions: create_session, get_token, introspect, and logout. A visitor gets a persistent anon@<uuid> identity plus a short-lived access token before they authenticate, with up to 1 KB of metadata attached at creation. The core is framework-agnostic RWA: it mounts no routes and sets no cookies. Tokens live in the integrator's store under a dedicated _a0_anon identifier, isolated from the authenticated _a0_session store. `AnonymousSession` never exposes the raw session token beyond what the caller already stored.
- Adds a token renewal ladder on get_token. A fresh cached access token is returned as-is. An expired access token is re-minted using the stored session token. An expired or invalid session token silently creates a brand-new session once. On that silent re-mint the metadata is lost and sub changes, and the returned `AnonymousSession` looks identical to a cached or re-minted one, so callers get no field-level signal that a new identity was issued. This path never raises, since a pre-login anonymous session carries no authorization.
- Injects the anonymous session_token into start_interactive_login() automatically when a session is active, sourced only from the SDK's own encrypted store and bound into `TransactionData` under the existing state binding. Injection is skipped when Pushed Authorization Requests are enabled, since the parameters go to the PAR endpoint instead.
- Adds typed anonymous option and response models (`AnonymousSession`, `AnonymousTokenResponse`, `AnonymousCreateTokenResponse`, `AnonymousSessionContext`, `AnonymousSessionIntrospection`) and a typed error hierarchy under `AnonymousSessionApiError`. The config subclasses (`AnonymousSessionFeatureNotEnabledError`, `AnonymousSessionClientNotEnabledError`, `AnonymousSessionClientNotSupportedError`, `AnonymousSessionResourceServerError`, `AnonymousSessionScopeError`) extend `AnonymousSessionCreateError`.
- Enforces metadata safeguards client-side before any network call: rejects dangerous keys (__proto__, constructor, prototype) with code invalid_metadata, and enforces a 1 KB UTF-8 JSON size cap with code metadata_too_large.

### Testing

Manual testing covered the following flows.

**Happy path**

1. Create - create_session mints an anon@<uuid> identity and returns a session_token plus an audience-bound access_token, both taken from the /anonymous/token response.
2. Get token - get_token walks the renewal ladder: cached, then re-mint via the session token, then a single silent new session.
3. Login injection - start_interactive_login() auto-injects the session_token into /authorize with no call-site change (skipped under PAR).
4. Logout - logout best-effort calls the server, then always clears the local store. A post-logout get_token raises `AnonymousSessionTokenError` (fail closed). Already-issued access tokens self-expire rather than being revoked.

**Negative / fail-closed**

| # | Scenario | Expected result | Status |
|---|----------|-----------------|--------|
| N1 | invalid audience | anonymous_resource_server_error | verified |
| N2 | ungranted scope | anonymous_scope_error | needs double checking |
| N3 | non-string metadata value | invalid_metadata (local, pre-network) | verified |
| N4 | dangerous key __proto__ | invalid_metadata | verified |
| N5 | metadata over 1 KB | metadata_too_large | verified |
| N6 | no session | `AnonymousSessionTokenError` | verified |

- [x] This change adds integration test coverage

### Checklist

### DIFF
--- a/src/auth0_server_python/auth_server/anonymous_client.py
+++ b/src/auth0_server_python/auth_server/anonymous_client.py
@@ -350,22 +350,27 @@ class AnonymousClient:
                 raise mapped
 
             try:
-                token_response = AnonymousCreateTokenResponse.model_validate(response.json())
-            except (json.JSONDecodeError, ValueError, ValidationError) as e:
+                body = response.json()
+            except (json.JSONDecodeError, ValueError) as e:
+                raise AnonymousSessionCreateError("Failed to parse anonymous token response") from e
+
+            if isinstance(body, dict) and not body.get("session_token"):
+                raise AnonymousSessionCreateError(
+                    "Anonymous token response contained no session_token. Enable session_token in the response for this tenant.",
+                    code="missing_session_token",
+                )
+
+            try:
+                token_response = AnonymousCreateTokenResponse.model_validate(body)
+            except (ValueError, ValidationError) as e:
                 raise AnonymousSessionCreateError("Failed to parse anonymous token response") from e
 
         now = int(time.time())
         context = AnonymousSessionContext(
             session_token=token_response.session_token,
-            sub=token_response.sub,
-            session_id=token_response.session_id,
             access_token=token_response.access_token,
             expires_at=now + token_response.expires_in,
-            session_expires_at=(
-                now + token_response.session_expires_in
-                if token_response.session_expires_in
-                else None
-            ),
+            session_expires_at=now + token_response.session_expires_in,
             metadata=metadata,
             created_at=now,
             domain=domain,
@@ -378,9 +383,8 @@ class AnonymousClient:
             options=store_options,
         )
         return AnonymousSession(
-            sub=context.sub,
-            session_id=context.session_id,
             access_token=context.access_token,
+            session_token=context.session_token,
             expires_at=context.expires_at,
             session_expires_at=context.session_expires_at,
             metadata=context.metadata,
@@ -416,6 +420,11 @@ class AnonymousClient:
         }
         if self._client_secret:
             body["client_secret"] = self._client_secret
+        # Re-mint must replay audience/scope. The endpoint doesn't remember them.
+        if context.audience:
+            body["audience"] = context.audience
+        if context.scope:
+            body["scope"] = context.scope
 
         async with self._get_http_client() as client:
             try:
@@ -449,19 +458,9 @@ class AnonymousClient:
                 if token_response.session_token is not None
                 else context.session_token
             ),
-            sub=token_response.sub if token_response.sub is not None else context.sub,
-            session_id=(
-                token_response.session_id
-                if token_response.session_id is not None
-                else context.session_id
-            ),
             access_token=token_response.access_token,
             expires_at=now + token_response.expires_in,
-            session_expires_at=(
-                now + token_response.session_expires_in
-                if token_response.session_expires_in
-                else context.session_expires_at
-            ),
+            session_expires_at=now + token_response.session_expires_in,
             metadata=context.metadata,
             created_at=context.created_at,
             domain=domain,
@@ -474,9 +473,8 @@ class AnonymousClient:
             options=store_options,
         )
         return AnonymousSession(
-            sub=new_context.sub,
-            session_id=new_context.session_id,
             access_token=new_context.access_token,
+            session_token=new_context.session_token,
             expires_at=new_context.expires_at,
             session_expires_at=new_context.session_expires_at,
             metadata=new_context.metadata,
@@ -603,7 +601,6 @@ class AnonymousClient:
         if context.domain and self._normalize_url(context.domain) != self._normalize_url(
             current_domain
         ):
-            # Never reuse a session across domains: discard and mint fresh.
             return await self._create_session_at(
                 current_domain,
                 audience=context.audience,
@@ -615,9 +612,8 @@ class AnonymousClient:
         now = int(time.time())
         if context.expires_at > now:
             return AnonymousSession(
-                sub=context.sub,
-                session_id=context.session_id,
                 access_token=context.access_token,
+                session_token=context.session_token,
                 expires_at=context.expires_at,
                 session_expires_at=context.session_expires_at,
                 metadata=context.metadata,
@@ -712,22 +708,21 @@ class AnonymousClient:
         if context is not None:
             domain = context.domain or await self._resolve_domain(store_options)
             base_url = f"https://{domain}"
-            body: dict[str, Any] = {
-                "client_id": self._client_id,
-                "session_token": context.session_token,
-            }
-            if self._client_secret:
-                body["client_secret"] = self._client_secret
+            auth = (
+                (self._client_id, self._client_secret) if self._client_secret else None
+            )
             try:
                 async with self._get_http_client() as client:
-                    response = await client.post(f"{base_url}/anonymous/logout", json=body)
+                    response = await client.post(
+                        f"{base_url}/anonymous/logout", json={}, auth=auth
+                    )
             except httpx.HTTPError as e:
                 error_to_raise = AnonymousSessionLogoutError(
                     "Failed to reach the anonymous logout endpoint"
                 )
                 error_cause = e
             else:
-                if response.status_code != 200:
+                if not 200 <= response.status_code < 300:
                     error_data = self._parse_anonymous_error_body(response)
                     mapped = self._map_anonymous_error(response.status_code, error_data, "logout")
                     if not isinstance(mapped, _AnonymousSessionExpired):
--- a/src/auth0_server_python/auth_server/anonymous_client.py
+++ b/src/auth0_server_python/auth_server/anonymous_client.py
@@ -12,6 +12,7 @@ from pydantic import ValidationError
 
 from auth0_server_python.auth_schemes.bearer_auth import BearerAuth
 from auth0_server_python.auth_types import (
+    AnonymousCreateTokenResponse,
     AnonymousSession,
     AnonymousSessionContext,
     AnonymousSessionIntrospection,
@@ -349,12 +350,9 @@ class AnonymousClient:
                 raise mapped
 
             try:
-                token_response = AnonymousTokenResponse.model_validate(response.json())
+                token_response = AnonymousCreateTokenResponse.model_validate(response.json())
             except (json.JSONDecodeError, ValueError, ValidationError) as e:
                 raise AnonymousSessionCreateError("Failed to parse anonymous token response") from e
-
-        if not token_response.session_token:
-            raise AnonymousSessionCreateError("Anonymous token response missing required fields")
 
         now = int(time.time())
         context = AnonymousSessionContext(
--- a/src/auth0_server_python/auth_server/anonymous_client.py
+++ b/src/auth0_server_python/auth_server/anonymous_client.py
@@ -199,7 +199,6 @@ class AnonymousClient:
 
         if code in ("session_expired", "invalid_session_token"):
             return _AnonymousSessionExpired(description)
-        # Distinguishes DPoP-mandated clients from a plain client-not-enabled block.
         if status_code == 400 and "Proof-of-Possession" in description:
             return AnonymousClientNotSupportedError(description, error_data)
         if code == "feature_not_enabled":
@@ -273,10 +272,6 @@ class AnonymousClient:
     def _decrypt_context(self, stored: Any) -> AnonymousSessionContext:
         """Decrypt and validate a stored anonymous session record.
 
-        A crypto-library failure and a validation failure both mean the
-        record is unusable, and both must convert to the same internal
-        signal instead of an untyped exception reaching the caller.
-
         Args:
             stored: The raw record read from the anonymous store.
 
@@ -313,8 +308,6 @@ class AnonymousClient:
     ) -> AnonymousSession:
         """Create a fresh anonymous session against a resolved domain.
 
-        Shared by create_session() and every renewal-ladder fallback.
-
         Args:
             domain: The resolved tenant domain.
             audience: Audience for the new session, or None.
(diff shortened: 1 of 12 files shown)


src/auth0_server_python/error/__init__.py
```diff
@@ -362,3 +362,95 @@ class PasskeyErrorCode:
     CHALLENGE_FAILED = "passkey_challenge_error"
     TOKEN_EXCHANGE_FAILED = "passkey_token_error"
     INVALID_RESPONSE = "invalid_response"
+
+
+# =============================================================================
+# Anonymous Session Error Classes
+# =============================================================================
+
+class AnonymousApiError(Auth0Error):
```
yogeshchoudhary147: **Spec deviation — error class names diverge from the requirements doc.**

The [SDK requirements doc](https://oktainc.atlassian.net/wiki/spaces/DXSDK/pages/1004601816) specifies:
```python
class AnonymousSessionError(Auth0Error): ...
class AnonymousSessionCreateError(AnonymousSessionError): ...
class AnonymousSessionTokenExpiredError(AnonymousSessionError): ...
```
This implementation ships `AnonymousApiError`, `AnonymousCreateError`, `AnonymousTokenError`, etc.

If the JS SDKs use the spec names, Python's public error surface will be inconsistent across SDKs. Any shared developer-facing error-handling documentation will show different class names per platform. If the rename is intentional, the spec should be updated to reflect it.

yogeshchoudhary147 (commented):
Reviewed against both the implementation and the [SDK requirements doc](https://oktainc.atlassian.net/wiki/spaces/DXSDK/pages/1004601816). Inline comments below cover spec deviations, confirmed bugs, and test gaps. Two earlier findings have been retracted: the `_normalize_url` str.replace concern (false positive for real-world domain inputs) and the claim that `httpx.HTTPError` catches 4xx/5xx responses (it does not — those are only raised via `raise_for_status()`, which `logout()` never calls).

src/auth0_server_python/error/__init__.py
```diff
@@ -362,3 +362,95 @@ class PasskeyErrorCode:
...
+
+
+class AnonymousCreateError(AnonymousApiError):
+    """Error thrown when creating or re-minting an anonymous session fails."""
+
+    def __init__(self, message: str, code: str = "anonymous_create_error", cause: Optional[dict] = None):
+        super().__init__(code, message, cause)
+
+
+class AnonymousLogoutError(AnonymousApiError):
```
yogeshchoudhary147: **`AnonymousLogoutError` is dead code — it can never be raised.**

`_map_anonymous_error()` maps `operation == "logout"` to this class, but `logout()` never calls `_map_anonymous_error()`. The `except httpx.HTTPError` in `logout()` catches only transport-level failures; non-2xx HTTP responses are silently ignored because the response object is never inspected at all. The only exception `logout()` can raise to a caller is `ConfigurationError` from `_require_store()`.

Either:
- `logout()` should check the response status, call `_map_anonymous_error()`, and re-raise on non-2xx (while still clearing local state), or
- `AnonymousLogoutError` and the `logout` branch in `_map_anonymous_error()` should be removed.

src/auth0_server_python/auth_server/anonymous_client.py
```diff
@@ -0,0 +1,711 @@
...
+    """
+    Client for Auth0 anonymous session operations.
+
+    Requires its own store instance, distinct from ServerClient's state_store.
+    DPoP is not supported with Anonymous Sessions.
+    """
+
+    def __init__(
+        self,
+        domain,
```
yogeshchoudhary147: **`domain` parameter is untyped.**

Every sibling client (`MfaClient`, `MyAccountClient`) types this as `Union[str, Callable]`. This is the only one that leaves it bare. Inconsistency will surface under type checkers and makes the parameter contract invisible to IDE users.

src/auth0_server_python/auth_server/anonymous_client.py
```diff
@@ -0,0 +1,711 @@
...
+        if metadata is None:
+            return
+        if not isinstance(metadata, dict):
+            raise AnonymousCreateError("metadata must be a JSON object", code="invalid_metadata")
+        for key, value in metadata.items():
+            if key in _DANGEROUS_METADATA_KEYS:
+                raise AnonymousCreateError(
+                    f"metadata key '{key}' is not allowed", code="invalid_metadata"
+                )
+            if not isinstance(value, str):
```
yogeshchoudhary147: **Metadata value restriction is narrower than the spec and should be confirmed against the actual API.**

The [requirements doc](https://oktainc.atlassian.net/wiki/spaces/DXSDK/pages/1004601816) defines `metadata` as `Record<string, unknown>` (any JSON value). This implementation rejects non-string values client-side. If the Auth0 API actually accepts non-string values, this check incorrectly blocks valid callers. If the API only accepts strings in practice, the spec is wrong and needs updating.

Consequently, `AnonymousSessionContext.metadata` is typed `Optional[dict[str, Any]]` (any value), while creation only permits `dict[str, str]`. The type annotation does not express the constraint, so the model's round-trip deserialization of a stored context containing an integer value would silently succeed at the Pydantic level.

src/auth0_server_python/auth_server/anonymous_client.py
```diff
@@ -0,0 +1,711 @@
...
+                raise mapped
+
+            try:
+                token_response = AnonymousTokenResponse.model_validate(response.json())
+            except (json.JSONDecodeError, ValueError, ValidationError) as e:
+                raise AnonymousTokenError("Failed to parse anonymous token response") from e
+
+        now = int(time.time())
+        new_context = AnonymousSessionContext(
+            session_token=token_response.session_token or context.session_token,
```
yogeshchoudhary147: **`or` idiom silently swallows empty strings for `Optional[str]` fields.**

```python
session_token=token_response.session_token or context.session_token,
sub=token_response.sub or context.sub,
session_id=token_response.session_id or context.session_id,
```

`"" or context.X` falls back to the stale context value, so if the API ever returns an empty string for any of these, the old value is silently persisted. Prefer explicit `None`-checks:
```python
session_token=token_response.session_token if token_response.session_token is not None else context.session_token,
```
This is consistent with how `session_expires_in` is handled two lines below.

src/auth0_server_python/auth_server/anonymous_client.py
```diff
@@ -0,0 +1,711 @@
...
+            domain = await self._resolve_domain(store_options)
+            return await self._create_session_at(
+                domain,
+                audience=self._default_audience,
+                scope=self._default_scope,
+                metadata=None,
+                store_options=store_options,
+            )
+
+        if self._domain_resolver:
```
yogeshchoudhary147: **Cross-tenant isolation gap in static-domain mode.**

The domain mismatch check is gated on `if self._domain_resolver:`. In static-domain mode `context.domain` is stored but never compared against `self._domain`. If an operator changes the configured domain (rare, but possible during a tenant migration) without clearing sessions, a session minted for the old domain is silently used against the new one. The isolation guarantee described in the guide does not hold unconditionally. Consider adding the same check for the static-domain case.

src/auth0_server_python/auth_server/anonymous_client.py
```diff
@@ -0,0 +1,711 @@
...
+                raise mapped
+
+            try:
+                return AnonymousSessionIntrospection.model_validate(response.json())
+            except (json.JSONDecodeError, ValueError, ValidationError) as e:
+                raise AnonymousIntrospectError(
+                    "Failed to parse anonymous introspection response"
+                ) from e
+
+    async def logout(self, store_options: Optional[dict[str, Any]] = None) -> None:
```
yogeshchoudhary147, line 677: **`logout()` never raises `AnonymousLogoutError` — see the comment on `error/__init__.py:392`.**

Additionally, there is no test covering the branch at line ~693 where `_decrypt_context` raises `_AnonymousSessionExpired` (sets `context = None` and skips the server call). That path clears local state correctly but is untested.

src/auth0_server_python/auth_types/__init__.py
```diff
@@ -852,3 +863,63 @@ class PasskeyTokenResponse(BaseModel):
...
+    metadata: Optional[dict[str, Any]] = None
+
+
+class AnonymousTokenResponse(BaseModel):
+    """Raw response from POST /anonymous/token."""
+
+    access_token: str
+    token_type: str = "Bearer"
+    expires_in: int
+    session_token: Optional[str] = None
```
yogeshchoudhary147, line 906: **`session_token` is `Optional` here but effectively required on the create path.**

`_create_session_at()` validates the response with this model and then immediately does:
```python
if not token_response.session_token:
    raise AnonymousCreateError("Anonymous token response missing required fields")
```

The `Optional` typing exists to accommodate the re-mint path (where the server may not return a new session token). Consider a separate narrow model for the create response, or at minimum a Pydantic validator that enforces presence, so the constraint is expressed in the type rather than in a manual post-validation check.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -551,6 +588,7 @@ async def start_interactive_login(
             domain=origin_domain,
             redirect_uri=auth_params.get("redirect_uri"),
             organization=resolved_org,
+            session_token=anonymous_session_token,
```
yogeshchoudhary147, line 664: **Question: should `complete_interactive_login()` promote `session_token` from `TransactionData` into `StateStore`?**

The auth0-server-js section of the [requirements doc](https://oktainc.atlassian.net/wiki/spaces/DXSDK/pages/1004601816) says:
> `completeInteractiveLogin()` — on callback, read the session token back from TransactionStore and promote it into StateStore as part of the authenticated session.

The Python section does not mention this step. The session token is saved into `TransactionData` here but nothing reads it back during the callback. If auth0-fastapi (GA) will need to reconstruct the anonymous session after login, this plumbing would need to exist in `auth0-server-python` first. Please confirm the omission is intentional for EA scope.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -8947,3 +8950,465 @@ async def test_complete_interactive_login_milliseconds_ceiling_fails_open(mocker
...
     stored_state = mock_state_store.set.call_args.args[1]
     assert stored_state.internal.session_expires_at is None
+
+
+# =============================================================================
+# ANONYMOUS SESSIONS - WIRING AND LOGIN-INJECTION TESTS
+# =============================================================================
+
+
+class _OneSlotStore:
```
yogeshchoudhary147: **`_OneSlotStore` is duplicated — identical class exists as `OneSlotStore` in `test_anonymous_client.py:41`.**

The explanatory comment about why `AsyncMock` is insufficient (identifier-as-salt, not location key) exists in both files. Moving it to `conftest.py` as a shared fixture would eliminate the duplication and keep the explanation in one place.

src/auth0_server_python/tests/test_anonymous_client.py
```diff
@@ -0,0 +1,739 @@
...
+)
+
+# Shared fixtures
+DOMAIN = "auth0.local"
+CLIENT_ID = "<client_id>"
+CLIENT_SECRET = "<client_secret>"
+SECRET = "test-secret-long-enough-for-encryption"
+
+
+class OneSlotStore:
```
yogeshchoudhary147: **Two test coverage gaps in `introspect()`:**

1. No test for what happens when the stored context is corrupted (invalid JWE) — `introspect()` should raise `AnonymousIntrospectError`, but this branch is untested. Compare to the equivalent test in `TestGetToken.test_corrupted_stored_token_triggers_silent_new_session`.
2. The `TestLogout` class has no test for the `context = None` branch (corrupted/missing context skips the server call but still clears local state).

src/auth0_server_python/error/__init__.py
rmad17: This was intentionally done by following the practice MFA error classes. 
MfaApiError's own action subclasses (MfaListAuthenticatorsError, MfaEnrollmentError, MfaChallengeError, MfaVerifyError) are flat — domain + action, no inserted noun.
However, on digging deeper found that there are exceptions like `MfaTokenExpiredError`/`MfaTokenInvalidError` where a noun - `Token` is added. Given that there is the presence of these exceptions I think it makes sense to keep the names consistent with JS names. 
Exception is the base error which will be `AnonymousSessionApiError` instead of `AnonymousSessionError` to maintain consistency with other base errors in the SDK. Will update the doc spec with this exception.

src/auth0_server_python/error/__init__.py
rmad17: `logout()` now checks `response.status_code`, maps non-2xx via `_map_anonymous_error()`, and raises `AnonymousSessionLogoutError` - but only after local state is cleared, so the session is always gone locally
  regardless of whether the remote call succeeded.

`session_expired`/`invalid_session_token` on the logout call itself are still swallowed. This isn't the SDK doc's renewal carve-out applied literally - that carve-out is paired with a remint, which logout has no equivalent of. It's swallowed because the goal state (no active anonymous session) is already true, so raising would flag a non-failure. Everything else surfaces per "surface all other errors... do not swallow them."

src/auth0_server_python/auth_server/anonymous_client.py
rmad17: Valid point.

src/auth0_server_python/auth_server/server_client.py
rmad17: This is out of EA scope and will be picked up when auth0-fastapi is worked upon.

src/auth0_server_python/auth_server/anonymous_client.py
rmad17: Both the points are addressed in [ee70a38](https://github.com/auth0/auth0-server-python/pull/156/changes/ee70a38fc116d6846ca32ae4150bc9d1f521072b) and https://github.com/auth0/auth0-server-python/pull/156/changes/15e47325a27b19f4e6ee8508e9aac429ce72dfe4

src/auth0_server_python/auth_types/__init__.py
rmad17: Addressed via https://github.com/auth0/auth0-server-python/pull/156/changes/6304af8c39f86af5086efe4597e6537af51fccba and https://github.com/auth0/auth0-server-python/pull/156/changes/61d727ebd39c9ad8a022aa6d0efb6c1a25330cb2

PR #158, opened 2026-08-19
feat: Private JWT CA support for passwordless

rmad17:
**Changed**

  `PasswordlessClient.start()`
  - Removed the hardcoded "client_secret": client._client_secret field from the request body.
  - Added client._apply_client_authentication(body, f"https://{origin_domain}/", in_body=True) after the body is built, so credentials are attached by the SDK's shared helper instead of inline.

  `PasswordlessClient.verify()`
  - Removed the same hardcoded "client_secret" field from the token-endpoint body.
  - Added client._apply_client_authentication(body, origin_issuer or f"https://{origin_domain}/", in_body=True), with the assertion audience taken from the discovery document's issuer and falling back to the
  resolved domain.

  **Testing**
  Tested via usual Passwordless flow as described in #153

### DIFF
--- a/src/auth0_server_python/auth_server/passwordless_client.py
+++ b/src/auth0_server_python/auth_server/passwordless_client.py
@@ -91,15 +91,16 @@ class PasswordlessClient:
             MissingRequiredArgumentError: When a magic link is requested but no
                 ``redirect_uri`` is configured on the client, or ``store_options``
                 is not provided.
+            ConfigurationError: When client authentication is not configured.
         """
         client = self._client
         origin_domain = await client._resolve_current_domain(store_options)
 
         body: dict[str, Any] = {
             "client_id": client._client_id,
-            "client_secret": client._client_secret,
             "connection": options.connection,
         }
+        client._apply_client_authentication(body, f"https://{origin_domain}/", in_body=True)
 
         if isinstance(options, StartPasswordlessEmailOptions):
             body["email"] = options.email
@@ -200,6 +201,7 @@ class PasswordlessClient:
             ApiError: When fetching the JWKS used to verify the ID token fails.
             SessionExpiredError: When the token's session-expiry ceiling is
                 already in the past.
+            ConfigurationError: When client authentication is not configured.
         """
         client = self._client
         origin_domain = await client._resolve_current_domain(store_options)
@@ -228,12 +230,12 @@ class PasswordlessClient:
         body: dict[str, Any] = {
             "grant_type": PASSWORDLESS_OTP_GRANT_TYPE,
             "client_id": client._client_id,
-            "client_secret": client._client_secret,
             "realm": options.connection,
             "username": options.username,
             "otp": options.verification_code,
             "scope": scope,
         }
+        client._apply_client_authentication(body, origin_issuer or f"https://{origin_domain}/", in_body=True)
         if options.audience:
             body["audience"] = options.audience
 
--- a/src/auth0_server_python/tests/test_passwordless_client.py
+++ b/src/auth0_server_python/tests/test_passwordless_client.py
@@ -6,6 +6,8 @@ from unittest.mock import AsyncMock, MagicMock
 
 import jwt
 import pytest
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
 from pydantic import ValidationError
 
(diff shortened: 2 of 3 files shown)


src/auth0_server_python/auth_server/passwordless_client.py
```diff
@@ -85,15 +85,16 @@ async def start(
...
         """
         client = self._client
         origin_domain = await client._resolve_current_domain(store_options)
 
         body: dict[str, Any] = {
             "client_id": client._client_id,
-            "client_secret": client._client_secret,
             "connection": options.connection,
         }
+        client._apply_client_authentication(body, f"https://{origin_domain}/", in_body=True)
```
kishore7snehil, line 97: Both this call and the one in verify() go through `_apply_client_authentication` unconditionally, so a client with no secret and no signing key now hits the `ConfigurationError` the helper raises. Passkey signin guards the same call with a check for a configured secret or key first, because it allows public clients. Before this change, start sent `client_secret` as None and still reached Auth0.

If passwordless is meant to always require a confidential client, this is the right behavior and worth a line in the docstring saying so. If public clients should still work here, shall we wrap the call the way passkey does? Either way it reads better as a deliberate choice than as a side effect of the helper.

kishore7snehil, line 97: The assertion audience here is built as `https://{origin_domain}/`, while verify() and the other token calls prefer the discovery issuer and fall back to that. start() has no discovery lookup so it can't read the issuer, and for a standard tenant the two come out the same. Just flagging the difference so it's a known choice.

src/auth0_server_python/tests/test_passwordless_client.py
```diff
@@ -1034,3 +1037,96 @@ async def test_verify_mfa_required_without_token_falls_through(self, mocker):
...
+            data["client_assertion"],
+            _public_key_pem(private_key),
+            algorithms=["RS256"],
+            audience=ISSUER,
+        )
+        assert claims["iss"] == CLIENT_ID
+        assert claims["sub"] == CLIENT_ID
+
+    @pytest.mark.asyncio
+    async def test_start_no_client_auth_configured_raises_configuration_error(self):
```
kishore7snehil, line 1166: This pins the no-auth-configured path for start(), but verify() takes the same new `ConfigurationError` path and has no equivalent test. The verify assertion test above also doesn't check that a secret-configured client puts `client_secret` in the body. This test is copyable for both.

kishore7snehil (changes requested):

src/auth0_server_python/tests/test_passwordless_client.py
```diff
@@ -1034,3 +1037,96 @@ async def test_verify_mfa_required_without_token_falls_through(self, mocker):
             )
         assert exc.value.code == "mfa_required"
         client._state_store.set.assert_not_awaited()
+
+
+# ── Private Key JWT (client assertion) client authentication ────────────────
+
+
+def _generate_rsa_private_key_pem() -> str:
```
kishore7snehil, line 1045: Optional, and not really about this PR: these two key helpers are copied from `test_server_client.py` where the same definitions already live. There's no shared test-helpers module yet, so this is the third copy floating around. Might be worth pulling them into a conftest fixture at some point so they don't drift.

kishore7snehil (changes requested):

src/auth0_server_python/auth_server/passwordless_client.py
rmad17: Auth0 docs mention that client secret is mandatory for RWA - https://auth0.com/docs/api/authentication/passwordless/authenticate-user#param-client-secret

Should we still add docstrings?

rmad17: Its a known choice. `auth0-server-js` currently takes a similar approach as well.

src/auth0_server_python/tests/test_passwordless_client.py
rmad17: Agreed, will create a JIRA ticket for this optimization.

src/auth0_server_python/auth_server/passwordless_client.py
kishore7snehil: It's optional

kishore7snehil (approved):
LGTM!

PR #159, opened 2026-08-21
feat: add mTLS (RFC 8705) client authentication

cschetan77:
## Summary

Adds Mutual TLS (RFC 8705) client authentication to `auth0-server-python`. When enabled, the SDK presents a TLS client certificate during the Auth0 token-endpoint handshake instead of a client secret — no credential travels in the request body.

- **Two new `ServerClient` constructor params:** `use_mtls: bool = False` and `ssl_context: Optional[ssl.SSLContext] = None`. The caller builds the `SSLContext` (`ssl.create_default_context()` + `load_cert_chain`); the SDK forwards it as `verify=ssl_context` to every `httpx.AsyncClient` it constructs (including the authlib client used for the authorization-code exchange).
- **Endpoint routing:** new `_resolve_token_endpoint(metadata)` helper returns `mtls_endpoint_aliases.token_endpoint` from the discovery document when mTLS is on, raising `ConfigurationError` if the alias is absent. All token-endpoint call sites in `server_client.py`, `mfa_client.py`, and `passwordless_client.py` are routed through it.
- **PAR routing:** new `_resolve_par_endpoint(metadata)` helper returns `mtls_endpoint_aliases.pushed_authorization_request_endpoint` when mTLS is on, so PAR requests also hit the mTLS alias.
- **Client authentication:** `_apply_client_authentication` returns `None` under mTLS — the certificate in the TLS handshake is the sole credential; no `client_secret` or `client_assertion` is added to the body. The authlib `AsyncOAuth2Client` is constructed with `client_secret=None` and `verify=ssl_context` under mTLS.
- **cnf.x5t#S256 warning:** `_warn_if_not_cert_bound` inspects the access token's `cnf` claim and emits a `logger.warning` when it is absent, signalling that Token Sender-Constraining is not active on the resource server. Scoped to `authorization_code` and `refresh_token` grants only (per RFC 8705 Table 1) — does not fire for MFA, passwordless, passkey, connection exchange, or custom token exchange, where the auth server intentionally omits `cnf`.
- **MFA under mTLS:** `MfaClient` receives `use_mtls` and `ssl_context` from `ServerClient` (cert on all MFA calls). A `token_endpoint_resolver` callable is injected at construction by `ServerClient` when mTLS is enabled — it calls `_resolve_current_domain` + `_get_oidc_metadata_cached` + `_resolve_token_endpoint`, keeping metadata caching in `ServerClient` where it belongs. `MfaClient` never fetches OIDC metadata directly. `/mfa/challenge` stays on the standard host; the client certificate is presented in the TLS handshake via `ssl_context`.
- **Passwordless under mTLS:** `/passwordless/start` stays on the standard host (no mTLS alias is advertised for it in the discovery doc; mTLS cert auth is not supported server-side on that endpoint). `passwordless.verify` routes the token exchange through the mTLS alias. Behaviour documented in `examples/MutualTLS.md`.
- **MyAccountClient under mTLS:** `ssl_context` is threaded through `MyAccountClient` so the client certificate is presented on `/me/` calls, allowing the resource server to verify `cnf.x5t#S256` binding.
- **DPoP mutual exclusion:** `signin_with_passkey` and `mfa.verify` raise `ConfigurationError` when both `dpop_key` and `use_mtls` are active — DPoP would bind the token to its own key and suppress `cnf.x5t#S256`, silently defeating mTLS token binding.
- **Passkey challenge/register under mTLS:** `/passkey/challenge` and `/passkey/register` are not served on the mTLS endpoint aliases, so a client using `use_mtls=True` (which forbids `client_secret`) cannot authenticate those endpoints. Documented in `examples/MutualTLS.md`.
- **Constructor validation** (all `ConfigurationError`, fail-fast): `use_mtls=True` without `ssl_context`; combined with `client_secret`; combined with `client_assertion_signing_key`.

## Changed files

| File | Change |
|------|--------|
| `auth_server/server_client.py` | `use_mtls`/`ssl_context` params, validation, `_resolve_token_endpoint`, `_resolve_par_endpoint`, `_apply_client_authentication` mTLS branch, `_warn_if_not_cert_bound` (cnf.x5t#S256 logger warning), DPoP guard, all token/PAR call-site routings, `_get_http_client` + `AsyncOAuth2Client` `verify=` injection |
| `auth_server/mfa_client.py` | `use_mtls`/`ssl_context` params, `token_endpoint_resolver` callable injection, `_get_http_client` `verify=` injection, DPoP guard in `verify()` |
| `auth_server/passwordless_client.py` | Token-endpoint call routed through `_resolve_token_endpoint` |
| `auth_server/my_account_client.py` | `ssl_context` threaded through for cert-bound token support on `/me/` calls |
| `tests/test_server_client.py` | Constructor validation, `_resolve_token_endpoint`, `_apply_client_authentication` mTLS branch, SSLContext threading, DPoP exclusion, alias-routing and body credential-absence assertions for interactive login, PAR, refresh token, backchannel, connection, and custom-exchange flows; `AsyncOAuth2Client` constructor verification; cnf warning tests |
| `tests/test_mfa_client.py` | SSLContext threading, `token_endpoint_resolver`, DPoP exclusion |
| `tests/test_passwordless_client.py` | `TestMtls` class: verify routing through mTLS alias, ssl_context threading on `start()` and `verify()`, `client_secret` absence assertions |
| `examples/MutualTLS.md` | New per-feature guide: prerequisites, cert generation, wiring, mutual-exclusion rules, cnf.x5t#S256 logger warning, MFA proxy note, passwordless caveat, passkey incompatibility callout |
| `README.md` | mTLS section linking to the guide and to Auth0 mTLS configuration docs |
| `references/flow-map.md` | mTLS row |

## Test plan

- [x] **Real-tenant integration testing** — validated end-to-end against a real Auth0 tenant with mTLS endpoint aliases enabled and a custom domain configured as a self-managed TLS proxy. A local nginx instance terminated TLS, extracted the client certificate, and forwarded it to Auth0's origin via the required headers — matching the production deployment model.

  Flows verified:
  - Discovery doc contains `mtls_endpoint_aliases` with `token_endpoint` resolving to the `mtls.` subdomain
  - Authorization code flow (login → callback → token exchange) routed to the mTLS alias
  - Access token contained `cnf.x5t#S256`; thumbprint matched the client certificate exactly
  - Refresh token flow (`grant_type=refresh_token`) routed through the mTLS alias; refreshed token retained `cnf.x5t#S256` with the same thumbprint
  - `get_user()` returned correct profile from the stored session

- [x] `poetry run pytest` — all 569 tests pass
- [x] `poetry run ruff check .` — no lint errors
- [x] Constructor raises `ConfigurationError` for: missing `ssl_context` with `use_mtls=True`; `client_secret` + `use_mtls`; `client_assertion_signing_key` + `use_mtls`
- [x] `_resolve_token_endpoint` returns the mTLS alias when present, raises `ConfigurationError` when alias is absent under mTLS, returns the standard endpoint when mTLS is off
- [x] `_resolve_par_endpoint` returns the mTLS PAR alias when `use_mtls=True`
- [x] `AsyncOAuth2Client` constructed with `client_secret=None` and `verify=ssl_context` under mTLS
- [x] `complete_interactive_login` under mTLS hits `mtls_endpoint_aliases.token_endpoint`
- [x] `get_token_by_refresh_token`, `backchannel_authentication_grant`, `get_token_for_connection`, and `custom_token_exchange` under mTLS all route to `mtls_endpoint_aliases.token_endpoint` and omit `client_secret` from the request body
- [x] `start_interactive_login` with PAR under mTLS hits `mtls_endpoint_aliases.pushed_authorization_request_endpoint` and omits `client_secret` from the request body
- [x] `passwordless.verify` under mTLS routes token exchange through the mTLS alias, omits `client_secret`, and presents `ssl_context`
- [x] `passwordless.start` under mTLS omits `client_secret` and passes `verify=ssl_context` to httpx
- [x] `_warn_if_not_cert_bound` emits `logger.warning` when `cnf.x5t#S256` is absent; silent on cert-bound tokens, opaque tokens, and missing access token
- [x] cnf warning fires on `complete_interactive_login` and `get_token_by_refresh_token` under mTLS; does not fire for MFA or other non-CNF grants
- [x] `signin_with_passkey` raises `ConfigurationError` when `dpop_key` + `use_mtls`
- [x] `mfa.verify` raises `ConfigurationError` when `dpop_key` + `use_mtls`; token endpoint resolved via injected `token_endpoint_resolver`


src/auth0_server_python/auth_server/server_client.py
```diff
@@ -270,6 +316,11 @@ def _apply_client_authentication(
         for reserved in ("client_secret", "client_assertion", "client_assertion_type"):
             params.pop(reserved, None)
 
+        if self._use_mtls:
+            # The client certificate presented in the TLS handshake is the sole
+            # credential; no body credential or HTTP basic auth is sent.
+            return None
```
kishore7snehil, line 361: This is where mTLS drops the body credential, and the alias resolver above does the endpoint routing. Both are applied on the interactive-login, refresh, backchannel, connection, custom-exchange, passkey, and MFA paths, but passwordless verify does not go through either. Under mTLS it still posts to the standard token endpoint and still puts `client_secret` in the body, and since a mTLS client has no secret, that call goes out with an empty credential to an endpoint that never sees the certificate, so the token it gets back is not certificate-bound.

Shall we route passwordless verify through the same resolver and drop its body credential when mTLS is on? If passwordless under mTLS is under consideration.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -241,15 +265,37 @@ def __init__(
             session_establisher=self._establish_session_from_mfa_verify_response,
             mfa_token_ttl=mfa_token_ttl,
             apply_client_authentication=self._apply_client_authentication,
+            use_mtls=self._use_mtls,
```
kishore7snehil, line 278: The My Account client built just above here does not get `use_mtls` or `ssl_context`, so under mTLS its API calls go out without the client certificate. The MFA client right here has both threaded in.

Shall we pass the same `use_mtls` and `ssl_context` through to the My Account client, or is My Account intentionally left off mTLS for now?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -765,6 +841,8 @@ async def complete_interactive_login(
             raise ApiError(
                 "token_error", f"Token exchange failed: {str(e)}", e)
 
+        self._warn_if_not_cert_bound(token_response.get("access_token"))
```
kishore7snehil: The certificate-bound warning fires here on interactive login and on the refresh path, but the backchannel, connection, custom-token-exchange, and passkey token responses all return without it.

Shall we call the warning on each path that returns a token, or funnel the token responses through one spot that always runs it?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -472,6 +479,7 @@ async def verify(
         options: dict[str, Any],
         store_options: Optional[dict[str, Any]] = None,
         dpop_key: Optional["jwk.JWK"] = None,
+        token_endpoint_override: Optional[str] = None,
```
kishore7snehil: The new `token_endpoint_override` param, and the `ConfigurationError` this now raises when `dpop_key` is combined with mTLS, are not in the docstring. The `Args` and `Raises` sections still describe the previous surface.

Worth adding both so the documented contract matches what the method does now.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -3290,12 +3370,18 @@ async def signin_with_passkey(
             raise MissingRequiredArgumentError("auth_session")
         if authn_response is None:
             raise MissingRequiredArgumentError("authn_response")
+        if self._use_mtls and dpop_key is not None:
```
kishore7snehil, line 3410: This now raises `ConfigurationError` when `dpop_key` is combined with mTLS, but the `Raises` section of the docstring still lists only the missing-argument, passkey, and organization errors.

We should add it there too.

examples/MutualTLS.md
```diff
@@ -0,0 +1,87 @@
...
+
+`use_mtls=True` cannot be combined with:
+
+| Parameter | Reason |
+|-----------|--------|
+| `client_secret` | One client-auth method only — Auth0 rejects requests carrying both. |
+| `client_assertion_signing_key` | Same — one method only. |
+| `dpop_key` (per-call on `signin_with_passkey` / `mfa.verify`) | DPoP binds to its own key (`cnf.jkt`) and suppresses `cnf.x5t#S256`; combining them silently defeats mTLS token binding. |
+
+All three raise `ConfigurationError` immediately (constructor for the first two, at the call site for DPoP).
```
kishore7snehil, line 58: One case that isn't covered here: `passkey_signup_challenge` and `passkey_login_challenge` hit `/passkey/register` and `/passkey/challenge`, which as far as I know are not served on the mTLS endpoint aliases, so they take a `client_secret` in the body rather than the certificate. A client configured for mTLS only has no `client_secret`, so those two calls would go out with just `client_id`.

Can you confirm against a tenant whether register and challenge actually work for a mTLS-only client? If they need a `client_secret`, it's worth calling that out here as the one exception, the way the rest of this doc is precise about what mTLS covers.

README.md
```diff
@@ -84,6 +84,29 @@ The key must be a PKCS8 PEM private key. Register its public key on your Auth0 a
...
+    client_id="<AUTH0_CLIENT_ID>",
+    use_mtls=True,
+    ssl_context=ssl_context,
+    secret="<AUTH0_SECRET>",
+)
+```
+
+`use_mtls=True` requires an Enterprise tenant with the Highly Regulated Identity add-on, a `self_managed_certs` custom domain, and mTLS endpoint aliases enabled. It cannot be combined with `client_secret`, `client_assertion_signing_key`, or a per-call `dpop_key` — each raises `ConfigurationError`.
+
+See [examples/MutualTLS.md](examples/MutualTLS.md) for the full setup guide, certificate generation, and token sender-constraining details.
```
kishore7snehil, line 108: The mTLS setup here lists the tenant prerequisites and Dashboard steps inline, but neither this section nor the guide links to the official Auth0 mTLS configuration doc. Since the tenant-side setup is canonical and maintained elsewhere, we can add a link to the Auth0 mTLS configuration docs alongside this reference.

src/auth0_server_python/tests/test_server_client.py
```diff
@@ -9677,3 +9678,221 @@ async def test_complete_interactive_login_milliseconds_ceiling_fails_open(mocker
     mock_state_store.set.assert_awaited_once()
     stored_state = mock_state_store.set.call_args.args[1]
     assert stored_state.internal.session_expires_at is None
+
+
+# ============================================================================
+# mTLS CLIENT AUTHENTICATION
```
kishore7snehil, line 9980: These mTLS tests are all grouped in one block at the end of the file. The interactive-login routing test and the passkey dpop-guard test each have a natural home next to that surface's existing tests, the way the Private Key JWT tests sit in their own section higher up.

Shall we move each test next to the surface it exercises, so each section shows what it covers?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -1382,7 +1460,7 @@ async def get_token_by_refresh_token(self, options: dict[str, Any]) -> dict[str,
             # Fetch OIDC metadata from the correct domain
             metadata = await self._get_oidc_metadata_cached(domain)
 
-            token_endpoint = metadata.get("token_endpoint")
+            token_endpoint = self._resolve_token_endpoint(metadata)
```
kishore7snehil, line 1497: The resolver has a unit test and interactive login has an end-to-end routing test, but the refresh, backchannel, connection, and custom-exchange paths that were also switched to the resolver have no test asserting they hit the mTLS alias. The resolver being right today does not stop a later edit to one of these call sites from quietly sending to the standard endpoint.

We should add a small routing assertion on each of the switched paths.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -3290,12 +3370,18 @@ async def signin_with_passkey(
             raise MissingRequiredArgumentError("auth_session")
         if authn_response is None:
             raise MissingRequiredArgumentError("authn_response")
+        if self._use_mtls and dpop_key is not None:
+            raise ConfigurationError(
+                "dpop_key cannot be combined with use_mtls. DPoP and mTLS bind tokens "
+                "differently; DPoP would take precedence and the token would not be "
```
kishore7snehil: The semicolon here splices two clauses. We can add two sentences which will increase the readibility.

README.md
```diff
@@ -84,6 +84,29 @@ The key must be a PKCS8 PEM private key. Register its public key on your Auth0 a
...
+auth0 = ServerClient(
+    domain="login.example.com",   # self_managed_certs custom domain
+    client_id="<AUTH0_CLIENT_ID>",
+    use_mtls=True,
+    ssl_context=ssl_context,
+    secret="<AUTH0_SECRET>",
+)
+```
+
+`use_mtls=True` requires an Enterprise tenant with the Highly Regulated Identity add-on, a `self_managed_certs` custom domain, and mTLS endpoint aliases enabled. It cannot be combined with `client_secret`, `client_assertion_signing_key`, or a per-call `dpop_key` — each raises `ConfigurationError`.
```
kishore7snehil: "or a per-call `dpop_key` — each raises" uses an em dash. A comma or a full stop will read the same.

examples/MutualTLS.md
```diff
@@ -0,0 +1,87 @@
...
+
+- Auth0 **Enterprise** tenant with the **Highly Regulated Identity** add-on
+- A `self_managed_certs` **custom domain** configured on the tenant
+- **Allow mTLS Endpoint Aliases** enabled on the tenant (Dashboard → Settings → Advanced)
+- Client application's authentication method set to **mTLS** in Dashboard → Applications → Settings → Credentials
+
+## Generating a client certificate (development)
+
+```bash
+# Self-signed CA + client cert (development only — use your PKI in production)
```
kishore7snehil: Em dash in the code comment: "development only — use your PKI in production".

examples/MutualTLS.md
```diff
@@ -0,0 +1,87 @@
...
+    ssl_context=ssl_context,
+    secret="<AUTH0_SECRET>",
+    authorization_params={
+        "audience": "<API_IDENTIFIER>",
+        "scope": "openid profile email offline_access",
+    },
+)
+```
+
+The SDK passes `ssl_context` as `verify=ssl_context` to every `httpx.AsyncClient` it constructs, including the authlib client used for the authorization-code exchange. You never call `load_cert_chain` inside the SDK — the caller owns the TLS material.
```
kishore7snehil: Em dash: "You never call `load_cert_chain` inside the SDK — the caller owns the TLS material".

examples/MutualTLS.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Mutual exclusion
+
+`use_mtls=True` cannot be combined with:
+
+| Parameter | Reason |
+|-----------|--------|
+| `client_secret` | One client-auth method only — Auth0 rejects requests carrying both. |
```
kishore7snehil: Em dash in the table cell: "One client-auth method only — Auth0 rejects requests carrying both".

examples/MutualTLS.md
```diff
@@ -0,0 +1,87 @@
...
+| `client_assertion_signing_key` | Same — one method only. |
```
kishore7snehil: Em dash: "Same — one method only".

examples/MutualTLS.md
```diff
@@ -0,0 +1,87 @@
...
+| `dpop_key` (per-call on `signin_with_passkey` / `mfa.verify`) | DPoP binds to its own key (`cnf.jkt`) and suppresses `cnf.x5t#S256`; combining them silently defeats mTLS token binding. |
+
+All three raise `ConfigurationError` immediately (constructor for the first two, at the call site for DPoP).
+
+## Token sender-constraining
+
+When the target API has **Token Sender-Constraining (mTLS)** enabled, issued access tokens carry a `cnf.x5t#S256` claim binding the token to the certificate thumbprint. The SDK warns if it receives a token that lacks this claim:
+
+> `UserWarning: mTLS is enabled but the access token is not certificate-bound (no cnf.x5t#S256). Sender-constraining is not active — configure Token Sender-Constraining (mTLS) on the API resource server.`
```
kishore7snehil: Em dash in the warning quote: "Sender-constraining is not active — configure Token Sender-Constraining".

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -393,6 +444,31 @@ async def _verify_and_decode_jwt(
...
+        if not self._use_mtls or not access_token:
+            return
+        try:
+            claims = jwt.decode(
+                access_token,
+                options={"verify_signature": False},
+                algorithms=["HS256", "RS256", "ES256", "PS256"],
+            )
+        except Exception:
+            return  # opaque or unparseable token — nothing to assert
```
kishore7snehil: Em dash in the inline comment: "opaque or unparseable token — nothing to assert".

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -393,6 +444,31 @@ async def _verify_and_decode_jwt(
...
+        cnf = claims.get("cnf") if isinstance(claims, dict) else None
+        if not (isinstance(cnf, dict) and cnf.get("x5t#S256")):
+            warnings.warn(
+                "mTLS is enabled but the access token is not certificate-bound "
+                "(no cnf.x5t#S256). Sender-constraining is not active — configure "
```
kishore7snehil: Em dash in the warning string: "Sender-constraining is not active — configure".

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -189,6 +193,25 @@ def __init__(
             self._domain = domain_str
             self._domain_resolver = None
 
+        self._use_mtls = use_mtls
+        self._ssl_context = ssl_context
+        if use_mtls:
+            if ssl_context is None:
+                raise ConfigurationError(
+                    "use_mtls=True requires an ssl_context with the client certificate "
+                    "loaded (ssl.create_default_context() + load_cert_chain())."
```
kishore7snehil: This look very cryptic. Can we change it to actual wordings?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -504,6 +512,12 @@ async def verify(
             MfaRequiredError: When chained MFA is required.
             ConfigurationError: If neither client_secret nor client_assertion_signing_key is configured.
         """
+        if self._use_mtls and dpop_key is not None:
+            raise ConfigurationError(
+                "dpop_key cannot be combined with use_mtls. DPoP and mTLS bind tokens "
+                "differently; DPoP would take precedence and the token would not be "
```
kishore7snehil: The semicolon here splices two clauses. We can add two sentences which will increase the readibility.

kishore7snehil (changes requested):

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -393,6 +444,31 @@ async def _verify_and_decode_jwt(
...
+            claims = jwt.decode(
+                access_token,
+                options={"verify_signature": False},
+                algorithms=["HS256", "RS256", "ES256", "PS256"],
+            )
+        except Exception:
+            return  # opaque or unparseable token — nothing to assert
+        cnf = claims.get("cnf") if isinstance(claims, dict) else None
+        if not (isinstance(cnf, dict) and cnf.get("x5t#S256")):
+            warnings.warn(
```
kishore7snehil: This is the only place the SDK calls `warnings.warn` for a runtime condition, and there's no other advisory of this kind in the codebase. The thing it flags is a setting on the API resource server rather than anything wrong at this call site, and the token that comes back is still valid and usable.

Is `UserWarning` the channel we want here, or would documenting in the docs the sender-constraining requirement and keeping this check out of the token path carry the same message without the once-only behavior? WDYT?

src/auth0_server_python/auth_server/mfa_client.py
```diff
@@ -534,7 +548,7 @@ async def verify(
             )
 
         try:
-            token_endpoint = f"{base_url}/oauth/token"
+            token_endpoint = token_endpoint_override or f"{base_url}/oauth/token"
```
rmad17: Under mTLS this falls back to the standard host because nothing ever passes `token_endpoint_override`, so MFA verify would fail with `invalid_client`. Could we resolve the alias inside `MfaClient` or via a `ServerClient` method?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -218,6 +241,7 @@ def __init__(
             client_id=client_id,
             client_secret=None if client_assertion_signing_key else client_secret,
             headers=self._telemetry_headers,
+            **({"verify": self._ssl_context} if self._use_mtls else {}),
         )
 
         self._my_account_client = MyAccountClient(
```
rmad17, line 255: The My Account client is built here, and `MyAccountClient` accepts no `use_mtls` or `ssl_context`, so it never attaches the certificate. If a token bound to the certificate is presented to a My Account resource server that enforces mTLS binding, the call would be rejected. Could you confirm whether My Account enforces binding?

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -752,7 +828,7 @@ async def complete_interactive_login(
         )
 
         try:
-            token_endpoint = self._oauth.metadata["token_endpoint"]
+            token_endpoint = self._resolve_token_endpoint(self._oauth.metadata)
```
rmad17, line 864: This path used to read `self._oauth.metadata["token_endpoint"]`, which raised `KeyError` when the endpoint was missing. It now calls `_resolve_token_endpoint`, which returns `None` when mTLS is off, and it does not add the `if not token_endpoint: raise` check the other five call sites carry. A missing endpoint would now pass `None` into `fetch_token` rather than failing early. Adding that check here would keep the behavior consistent.

src/auth0_server_python/auth_server/server_client.py
```diff
@@ -241,15 +265,37 @@ def __init__(
...
         self._passwordless_client = PasswordlessClient(self)
 
     def _get_http_client(self, **kwargs) -> httpx.AsyncClient:
         """Return an httpx.AsyncClient with telemetry headers injected."""
         headers = {**kwargs.pop("headers", {}), **self._telemetry_headers}
+        if self._use_mtls and "verify" not in kwargs:
+            kwargs["verify"] = self._ssl_context
         return httpx.AsyncClient(headers=headers, **kwargs)
 
+    def _resolve_token_endpoint(self, metadata: dict) -> Optional[str]:
```
rmad17, line 298: This resolves the mTLS alias for the token endpoint only. The PAR path reads `pushed_authorization_request_endpoint` straight from standard metadata, so PAR plus mTLS posts to the standard host and would hit the same `invalid_client`. Is PAR to be supported in this flow? If yes, could the alias resolution here be extended to cover the PAR endpoint, or PAR fail closed when both are enabled until it is?

rmad17 (changes requested):

src/auth0_server_python/auth_server/server_client.py
cschetan77: Fixed. `passwordless_client.py` now routes the token endpoint call through `_resolve_token_endpoint(metadata)`, the same resolver as all other token call sites, so under mTLS it hits
  `mtls_endpoint_aliases.token_endpoint`. The credential drop was already handled: `_apply_client_authentication` returns `None` under mTLS and was already being called at that call site. Only
  the endpoint routing needed updating.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Fixed. `ssl_context` is now threaded through `MyAccountClient` so the client certificate is presented on `/me/` calls.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Decided to remove `_warn_if_not_cert_bound` entirely rather than adding it to every token path. The warning flagged a resource server configuration issue the SDK has no control over. Instead `MutualTLS.md` now documents the `cnf.x5t#S256` requirement directly under "Token sender-constraining". 
Let me know if you think otherwise, will implement something to funnel the token responses through one spot that always runs it.

src/auth0_server_python/auth_server/mfa_client.py
cschetan77: With new commits,`token_endpoint_override` was replaced entirely by a `token_endpoint_resolver` callable injected at construction, it's an internal detail of how `ServerClient` wires `MfaClient`, so nothing needed in the public docstring for it. The `ConfigurationError` for `dpop_key` + `use_mtls` is now in the `Raises` section.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Added. The `Raises` section now includes `ConfigurationError` for the `dpop_key` + `use_mtls` combination.

examples/MutualTLS.md
cschetan77: Confirmed, `/passkey/register` and `/passkey/challenge` are not served on the mTLS endpoint aliases. The [auth0-auth-js](https://github.com/auth0/auth0-auth-js/blob/34578d4c6dacac3cebeb79b1c9f75c7221b66790/packages/auth0-auth-js/src/passkey/types.ts#L45) SDK comments this explicitly: a client relying on mTLS has no credential for them and is rejected once the certificate headers are absent. `MutualTLS.md` now has a dedicated callout explaining this constraint. We documented it rather than raise at runtime since the constraint is Auth0 server topology, not SDK logic, and it matches the JS SDK's approach.

README.md
cschetan77: Added: https://auth0.com/docs/get-started/applications/configure-mtls, this page covers the tenant-side application configuration described in the README section.

src/auth0_server_python/tests/test_server_client.py
cschetan77: Done. `test_signin_with_passkey_rejects_dpop_under_mtls` moved into the PASSKEY AUTHENTICATION section, `test_complete_interactive_login_uses_mtls_token_endpoint` moved into the IPSIE section.
The remaining mTLS tests (constructor validation, resolver unit tests, credential-drop) stay in the mTLS section since they have no specific flow to live next to. This matches the Private Key JWT pattern, auth-method-specific tests in their own section, flow-level tests next to their flow.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Added four tests, each placed next to the flow it exercises:

  - `test_get_token_by_refresh_token_uses_mtls_token_endpoint`
  - `test_backchannel_authentication_grant_uses_mtls_token_endpoint`
  - `test_get_token_for_connection_uses_mtls_token_endpoint`
  - `test_custom_token_exchange_uses_mtls_token_endpoint`

  Each mocks `_get_oidc_metadata_cached` to return metadata with `mtls_endpoint_aliases.token_endpoint` and asserts the HTTP POST hits the alias URL, not the standard endpoint.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Reworded to plain English. The error message now reads:

  ```
  ssl_context must be a valid ssl.SSLContext object. Build one with ssl.create_default_context() and call load_cert_chain() on it before passing it in.
  ```

src/auth0_server_python/auth_server/server_client.py
cschetan77: Agreed. Removed `_warn_if_not_cert_bound` entirely. 
Summing up the reasons (mentioned by you as well):
  1. It was the only `warnings.warn` call in the SDK, inconsistent with the rest of the codebase.
  2. It flagged a resource server configuration issue the SDK has no control over.
  3. Python's default warning deduplication means it fires once then goes silent, giving false confidence it had been seen.
  4. The token is valid and usable regardless.

`MutualTLS.md` now documents the `cnf.x5t#S256` requirement directly under "Token sender-constraining", which seems to be the right place for a server-configuration note.

src/auth0_server_python/auth_server/mfa_client.py
cschetan77: Fixed. `token_endpoint_override` was probably the wrong solution, nothing was ever passing it. Replaced it with a `token_endpoint_resolver` callable injected at `MfaClient` construction time.
`ServerClient` passes `self._resolve_mfa_token_endpoint`, an async callable that receives `store_options` and chains:
  ```
  _resolve_current_domain → _get_oidc_metadata_cached → _resolve_token_endpoint
  ```

`MfaClient.verify()` calls it when the resolver is set. This keeps metadata caching in `ServerClient` where it belongs and keeps the framework call (`client.mfa.verify(options, store_options)`) clean with no extra arguments.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Fixed. `ssl_context` is now wired through `MyAccountClient` so the client certificate accompanies every `/me/` call. 

> Could you confirm whether My Account enforces binding? - Confirmed

src/auth0_server_python/auth_server/server_client.py
cschetan77: Fixed. Added `if not token_endpoint: raise ApiError(...)` after `_resolve_token_endpoint` in `complete_interactive_login`.

Before this PR, `self._oauth.metadata["token_endpoint"]` raised `KeyError` on a missing endpoint. After switching to `_resolve_token_endpoint`, a missing non-mTLS endpoint returns `None` and would have passed silently into `fetch_token`. The other five call sites already carried this guard, `complete_interactive_login` was the one gap.

src/auth0_server_python/auth_server/server_client.py
cschetan77: Fixed. Added `_resolve_par_endpoint(metadata)` which returns `mtls_endpoint_aliases.pushed_authorization_request_endpoint` under mTLS, mirroring `_resolve_token_endpoint`. The PAR request in `start_interactive_login` now routes through this resolver.

A routing assertion test (`test_par_request_uses_mtls_alias_endpoint`) is included.

cschetan77:
All style fixes addressed in one commit (`style: apply repo conventions to mTLS code and docs`):

  - **Semicolon splices** (×2): split into two sentences in both `server_client.py` and `mfa_client.py`.
  - **Em dashes in `README.md` and `MutualTLS.md`** (×5): replaced with plain hyphens throughout.
  - **Em dashes in `server_client.py`** (×2): the inline comment and warning string were part of `_warn_if_not_cert_bound`, which was removed entirely — the em dashes are gone with it.
  - **Em dash in `MutualTLS.md` warning quote** (×1): the quoted warning text was removed along with `_warn_if_not_cert_bound`; `MutualTLS.md` now documents the requirement directly.

Piyush-85 (dismissed):
LGTM

kishore7snehil (dismissed):
LGTM!

src/auth0_server_python/auth_server/server_client.py
cschetan77: [c3cf72c](https://github.com/auth0/auth0-server-python/pull/159/commits/c3cf72ca856c0d6f8dc0600a4c5f334f2ddc9240)

rmad17 (approved):
LGTM!

PR #161, opened 2026-08-26
docs: Adds Step-up authentication example

kailash-b:
### Changes

Adds a new example guide, **`examples/StepUpAuthentication.md`**, documenting how to perform step-up authentication (MFA) with `auth0-server-python`.

This is **documentation only** — no SDK source, public API, or behavior changes. The guide explains how integrators can require a stronger authentication factor before a sensitive action while keeping the user's existing session.

### Testing

- [ ] This change adds unit test coverage
- [ ] This change adds integration test coverage
- [ ] This change has been tested on the latest version of the platform/language or why not

### Checklist

### DIFF
--- a/examples/StepUpAuthentication.md
+++ b/examples/StepUpAuthentication.md
@@ -11,7 +11,7 @@ For the concepts, tenant requirements, and available `acr_values` policies, see 
 
 ## Requesting Step-Up
 
-Pass `acr_values` inside `authorization_params` when starting the login. Add `max_age: 0` to force a fresh authentication instead of silently reusing the current session, and use `app_state` to remember where to resume.
+Pass `acr_values` inside `authorization_params` when starting the login. Add `max_age: 0` to force a fresh authentication instead of silently reusing the current session, and use `app_state` to remember where to resume. This assumes a configured `server_client` (see the constructor in [InteractiveLogin.md](./InteractiveLogin.md), or the full example below):
 
 ```python
 from auth0_server_python.auth_types import StartInteractiveLoginOptions
--- a/examples/StepUpAuthentication.md
+++ b/examples/StepUpAuthentication.md
@@ -1,29 +1,29 @@
 # Step-Up Authentication (MFA)
 
-Step-up authentication asks an already-logged-in user to re-authenticate with a stronger factor before a sensitive action (transferring funds, changing security settings, deleting data). The user keeps their session; you just require higher assurance for the one action.
+Step-up authentication asks an already-logged-in user to re-authenticate with a stronger factor before a sensitive action (transferring funds, changing security settings, deleting data). The user keeps their session. You just require higher assurance for the one action.
 
-With `auth0-server-python` you request step-up through the ordinary redirect flow — the same `start_interactive_login()` / `complete_interactive_login()` pair from [InteractiveLogin.md](./InteractiveLogin.md) — by adding an `acr_values` authorization parameter. Auth0's Universal Login handles the MFA challenge; when the user returns, you confirm it by reading the `acr` / `amr` claims from the session.
+With `auth0-server-python` you request step-up through the ordinary redirect flow - the same `start_interactive_login()` / `complete_interactive_login()` pair from [InteractiveLogin.md](./InteractiveLogin.md) - by adding an `acr_values` authorization parameter. Auth0's Universal Login handles the MFA challenge. When the user returns, you confirm it by reading the `acr` / `amr` claims from the session.
 
-For the concepts, tenant requirements, and available `acr_values` policies, see the [Auth0 step-up authentication docs](https://auth0.com/docs/secure/multi-factor-authentication/step-up-authentication).
+For the concepts, tenant requirements, and available `acr_values` policies, see the [Auth0 step-up authentication for web apps docs](https://auth0.com/docs/secure/multi-factor-authentication/step-up-authentication/configure-step-up-authentication-for-web-apps).
 
 > [!NOTE]
-> This is the right approach for apps using the **hosted login (Universal Login) redirect flow**. If your app drives an **embedded** MFA challenge itself (no redirect), use the MFA API via `client.mfa` instead — see [MFA.md](./MFA.md).
+> This is the right approach for apps using the **hosted login (Universal Login) redirect flow**. If your app drives an **embedded** MFA challenge itself (no redirect), use the MFA API via `client.mfa` instead - see [MFA.md](./MFA.md).
 
 ## Requesting Step-Up
 
-Pass `acr_values` inside `authorization_params` when starting the login. Add `max_age: 0` to force a fresh authentication instead of silently reusing the current session, and use `app_state` to remember where to resume:
+Pass `acr_values` inside `authorization_params` when starting the login. Add `max_age: 0` to force a fresh authentication instead of silently reusing the current session, and use `app_state` to remember where to resume.
 
 ```python
 from auth0_server_python.auth_types import StartInteractiveLoginOptions
 
-# The OIDC "multi-factor" policy URI — asks Auth0 to enforce MFA for this login.
+# The OIDC "multi-factor" policy URI - asks Auth0 to enforce MFA for this login.
 MFA_ACR = "http://schemas.openid.net/pape/policies/2007/06/multi-factor"
 
 authorization_url = await server_client.start_interactive_login(
     StartInteractiveLoginOptions(
         authorization_params={
             "acr_values": MFA_ACR,
-            "max_age": 0,  # force re-authentication; do not reuse the existing session
+            "max_age": 0,  # force a fresh authentication instead of reusing the session
         },
         app_state={"returnTo": "/transfer"},  # resume this action after step-up
     ),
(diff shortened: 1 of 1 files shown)


examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
+# Step-Up Authentication (MFA)
+
+Step-up authentication asks an already-logged-in user to re-authenticate with a stronger factor before a sensitive action (transferring funds, changing security settings, deleting data). The user keeps their session; you just require higher assurance for the one action.
+
+With `auth0-server-python` you request step-up through the ordinary redirect flow — the same `start_interactive_login()` / `complete_interactive_login()` pair from [InteractiveLogin.md](./InteractiveLogin.md) — by adding an `acr_values` authorization parameter. Auth0's Universal Login handles the MFA challenge; when the user returns, you confirm it by reading the `acr` / `amr` claims from the session.
+
+For the concepts, tenant requirements, and available `acr_values` policies, see the [Auth0 step-up authentication docs](https://auth0.com/docs/secure/multi-factor-authentication/step-up-authentication).
```
kishore7snehil: This link points to Auth0's general step-up page, which teaches step-up through scopes (requesting a high-value scope triggers MFA), not the `acr_values` policy approach this guide actually uses. It never mentions `acr_values`, `acr`, or `amr`, so a reader who follows it to learn more lands on a different method than the one they just copied.

Auth0 has a web-app specific page that describes exactly this flow, requesting `acr_values` with the multi-factor policy URI and reading the `acr` / `amr` claims. Shall we point the link there so the guide and its reference match?

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+
+```python
+from typing import Optional
+
+MFA_ACR = "http://schemas.openid.net/pape/policies/2007/06/multi-factor"
+
+def mfa_completed(user: Optional[dict]) -> bool:
+    if not user:
+        return False
+    return user.get("acr") == MFA_ACR or "mfa" in (user.get("amr") or [])
```
kishore7snehil, line 47: Once a user steps up, `acr` and `amr` stay on the session for its whole lifetime, so this reads as true for every later sensitive action, not just the one that triggered the step-up. The guide frames step-up as re-authenticating before a sensitive action, but as written the check is per-session, so a user who stepped up to transfer funds an hour ago passes the gate on the next sensitive action with no fresh prompt.

Can we call this out, and if a fresh challenge per action is the intent, show checking login recency (e.g. the `auth_time` claim) rather than just the presence of `acr` / `amr`?

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+
+    result = await server_client.complete_interactive_login(
+        callback_url, store_options=store_options
+    )
+    # Return where to send the user next; your app performs the redirect.
+    return (result.get("app_state") or {}).get("returnTo", "/")
+```
+
+> [!NOTE]
+> The redirect itself is framework-specific — these handlers return the URL to redirect to, and your app issues the actual HTTP redirect (e.g. a `302`/`303`). `max_age: 0` matters: without it, a user who authenticated moments ago may be returned straight to your callback without a fresh MFA prompt.
```
kishore7snehil: This NOTE covers `max_age`, but there's no mention that the tenant has to be configured to enforce MFA for the requested policy. If it isn't, the user returns without `acr` / `amr`, `mfa_completed` stays false, and `handle_transfer` sends them to step up again, so they loop.

Shall we add a line noting the tenant needs the MFA policy set up, and that the app should avoid re-redirecting when the claim still isn't present after a return?

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+    if not user:
+        return False
+    return user.get("acr") == MFA_ACR or "mfa" in (user.get("amr") or [])
+```
+
+`acr` and `amr` are standard OIDC claims the SDK surfaces on the user object alongside `sub`, `email`, and the rest. They are only present when the corresponding authentication took place, so a plain login (no step-up) will not carry them.
+
+## Full Example: Gating a Sensitive Action
+
+`auth0-server-python` is framework-agnostic — it only reads and writes through the `request` / `response` objects you hand it via `store_options`. The handlers below are plain `async` functions; wire them into whatever framework you use. They gate a **Transfer Funds** action: if the session has not completed MFA, they return the authorization URL for your app to redirect to, then resume the transfer when the user returns.
```
kishore7snehil: Keeping these handlers framework-agnostic reads well for the core SDK, and calling that out here is the right move. The one thing a reader still has to work out is how to turn the returned URL into an actual redirect.

Shall we add a short "in FastAPI this looks like" snippet for just the redirect step, while keeping the full runnable app in the framework guide (auth0-fastapi) where it fits better?

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+        return await server_client.start_interactive_login(
+            StartInteractiveLoginOptions(
+                authorization_params={"acr_values": MFA_ACR, "max_age": 0},
+                app_state={"returnTo": "/transfer"},
+            ),
+            store_options=store_options,
+        )
+
+    # MFA satisfied for this session — safe to run the sensitive action.
+    run_transfer()
```
kishore7snehil: The block titled "Full Example" calls `run_transfer()`, but nothing in the doc defines or imports it, so a reader who copies the whole block hits a `NameError` on the first run. The `request` / `response` names read clearly as framework stand-ins, but a bare `run_transfer()` looks like real SDK usage.

Shall we add a one-line stub or a comment marking it as your own transfer logic?

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+
+Pass `acr_values` inside `authorization_params` when starting the login. Add `max_age: 0` to force a fresh authentication instead of silently reusing the current session, and use `app_state` to remember where to resume:
+
+```python
+from auth0_server_python.auth_types import StartInteractiveLoginOptions
+
+# The OIDC "multi-factor" policy URI — asks Auth0 to enforce MFA for this login.
+MFA_ACR = "http://schemas.openid.net/pape/policies/2007/06/multi-factor"
+
+authorization_url = await server_client.start_interactive_login(
```
kishore7snehil, line 22: This snippet uses `server_client` before it's constructed anywhere in the guide. The client is only built lower down, in the full example. The interactive-login guide sets the client up first and then calls it.

Shall we add the short constructor here too, or a line pointing to where it's created, so this block stands on its own?

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
+# Step-Up Authentication (MFA)
+
+Step-up authentication asks an already-logged-in user to re-authenticate with a stronger factor before a sensitive action (transferring funds, changing security settings, deleting data). The user keeps their session; you just require higher assurance for the one action.
```
kishore7snehil: Two full clauses spliced with a semicolon here. Reads cleaner as two sentences:

"The user keeps their session. You just require higher assurance for the one action."

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+
+With `auth0-server-python` you request step-up through the ordinary redirect flow — the same `start_interactive_login()` / `complete_interactive_login()` pair from [InteractiveLogin.md](./InteractiveLogin.md) — by adding an `acr_values` authorization parameter. Auth0's Universal Login handles the MFA challenge; when the user returns, you confirm it by reading the `acr` / `amr` claims from the session.
```
kishore7snehil: Same semicolon splice. As two sentences:

"Auth0's Universal Login handles the MFA challenge. When the user returns, you confirm it by reading the `acr` / `amr` claims from the session."

examples/StepUpAuthentication.md
```diff
@@ -0,0 +1,114 @@
...
+
+For the concepts, tenant requirements, and available `acr_values` policies, see the [Auth0 step-up authentication docs](https://auth0.com/docs/secure/multi-factor-authentication/step-up-authentication).
+
+> [!NOTE]
+> This is the right approach for apps using the **hosted login (Universal Login) redirect flow**. If your app drives an **embedded** MFA challenge itself (no redirect), use the MFA API via `client.mfa` instead — see [MFA.md](./MFA.md).
```
kishore7snehil: A few lines use em dashes (this note, the intro, the framework-agnostic paragraph, and the closing note). Plain hyphens read fine here if you want to keep the punctuation simple.

kishore7snehil (changes requested):

examples/StepUpAuthentication.md
kailash-b: Addressed in 5aa9110247047613b004f90a4de290e61299c92c
Points to the correct URL

examples/StepUpAuthentication.md
kailash-b: Added a callout that this is per-session and what the users can do if they require a fresh action per challenge. 
Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Added a small fastpi snippet towards the end. 
Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Added a comment.
Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Addressed in 5aa9110247047613b004f90a4de290e61299c92c

examples/StepUpAuthentication.md
kailash-b: Addressed

kishore7snehil (approved):
LGTM!

PR #170, opened 2026-09-10
Release v1.0.0b16

rmad17:
**Added**
- feat: add mTLS (RFC 8705) client authentication - new `use_mtls` and `ssl_context` params let `ServerClient` present a TLS client certificate at the token endpoint instead of sending a client secret [\#159](https://github.com/auth0/auth0-server-python/pull/159) ([cschetan77](https://github.com/cschetan77))


kishore7snehil (approved):
LGTM!

